PR #4, opened 2023-07-03
Implement build-debian action

EduardGomezEscandell:
This action builds a debian package at a specified path and uploads the artifacts generated. It is a wrapper around jtdor/build-deb-action, which itself uses dpkg.

It needs:
- the source dir: where the debian/ folder is located
- docker-image: where the package will be built, and as such the release to target
- token: This one is optional. If provided, it is used to grant git  access to private repos, which is necessary to fetch Go private packages

This action has been tested and is in use in a private repository of the Canonical organization.

UDENG-878

### DIFF
--- a/gh-actions/common/build-debian/action.yml
+++ b/gh-actions/common/build-debian/action.yml
@@ -1,0 +1,113 @@
+name: Build debian package
+description: Builds a debian package and uploads the artifact
+
+inputs:
+  source-dir:
+    required: false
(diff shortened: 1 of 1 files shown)


EduardGomezEscandell:
CC @didrocks

didrocks (approved):
Excellent work! Great that you didn’t have to change it compared our first pass in the wsl repo :)

Time now to move all our projects to that generic workflow!

PR #6, opened 2023-09-21
fix(github): Ensure correct group of people is set for reviews

didrocks:
The codeowner file syntax is actually overriding previous rules and is not cumulative. We can assign owners for subparts of the trees, but without rule, this is a global override.
Then, ensure that both owners are assigned for review on this repository.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,1 @@
-* @didrocks
-* @tim-hm
+* @didrocks @tim-hm


tim-hm:
Oh how interesting ... i always assume it was cumulative so I'll need to changed it elsewhere. Nice catch!

tim-hm (approved):

didrocks:
> Oh how interesting ... i always assume it was cumulative so I'll need to changed it elsewhere. Nice catch!

I thought that too, until last week’s Amin’s PR, I noted it down to look at it when I got some minutes, which was… yesterday :p

PR #8, opened 2023-09-29
feat(gh-actions): allow action caller to specify a working directory

EduardGomezEscandell:
This is useful (needed really) if you need to run actions on a sub-directory of a monorepo.

### DIFF
--- a/gh-actions/common/has-diff/action.yml
+++ b/gh-actions/common/has-diff/action.yml
@@ -2,6 +2,9 @@ name: Has diff
 description: Returns if there is diff within the repository from the current directory.
 
 inputs:
+  working-directory:
+    description: Directory were to run the action. All other paths are relative to this one.
+    default: '.'
   regexp-to-ignore:
     description: Regexp patterns passed to ignore-matching-lines when diffing.
   paths-to-ignore:
@@ -20,6 +23,7 @@ runs:
     - name: Returns if there is a local diff
       id: has-diff
       shell: bash
+      working-directory: ${{ inputs.working-directory }}
       run: |
         echo Returns if there is a local diff
         set -eu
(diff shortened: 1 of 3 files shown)


EduardGomezEscandell:
Pinging @didrocks for a sanity check.

didrocks (changes requested):
I think that rather than duplicating working-directory in most of the jobs, we can use a default that is overridable per job, like in https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iddefaultsrun. WDYT?

EduardGomezEscandell:
> it seems this is only for run jobs, so won’t fix all the cases, maybe digging into that direction to see if anything exists?

Yeah, that is the problem. I did digg a bit and I couldn't find any way to change the working dir (otherwise this PR would not be necessary, as the caller could use such a tool).

The best we can do is use the default for run, and otherwise be careful with every other action (pre-prend to paths, see if they have their own working-dir option, etc).

didrocks:
> The best we can do is use the default for run, and otherwise be careful with every other action (pre-prend to paths, see if they have their own working-dir option, etc).

+1

EduardGomezEscandell:
Well, it seemse that default values for jobs only work in Workflows, not actions. So it seems like this PR is ready.

didrocks:
You can have it either on the workflow but also at the job level: https://docs.github.com/en/actions/using-jobs/setting-default-values-for-jobs#setting-default-values-for-a-specific-job. Did you try the job level too in the action?

EduardGomezEscandell:
Actions don't have jobs, only `runs`. The following yaml is not valid:
```yaml
runs:
  uses: ubuntu-latest
  defaults:
    run:
      working-directory: ${{ inputs.working-directory }}
  steps:
    # ...
```

In general, the schema for actions is much more limited:
https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-composite-actions

didrocks:
Ack then, let’s go with this! Thanks for digging :)

didrocks (approved):
Thanks for looking for the simplification! Unfortunate that we couldn't though.

I’ll just update the merge commit title before merging. Thanks!

PR #9, opened 2023-10-02
fix(gh-actions): adapt workaround for golangci-lint on mono-repo

EduardGomezEscandell:
Follow-up to #8


didrocks:
> Sure! Is there a guide for the commit naming convetion?

See the pin comment in the MM channel as we are experimenting it. Once this is working well for a couple of weeks, the plan is to integrate those guidances into this repo.

didrocks (approved):
I did retitle the PR to describe what’s the intent is. Otherwise, good!

PR #10, opened 2023-10-02
fix(gh-actions): Fix Windows issues in cross-platform Github actions

EduardGomezEscandell:
Strictly necessary:
- (85a1bbc) `sudo` is not available on Windows.
- (e3c87d8) `rev` is not available on Windows. I switched to using jq to parse go mod files.
- (5996fb2) Treat Linux-specific workarounds with more finesse.
- (d88da61, 9068ca3) Windows path arguments are passed `\with\back\slashes` which are read by bash as `withbackslashes`, making all usage of `cd` fail.

A couple QoL things:
- (2c5235b) Use single quotes around '${{ github.variable }}' to avoid weird interactions with special characters (were possible). Github pretty much acts like the C pre-processor here.
- (e5d5938) Added a ::group:: were I saw `apt` spam.

---
UDENG-1399

PR #11, opened 2023-10-05
fix(gh-actions): Fix bash variable name in has-diff action

EduardGomezEscandell:
Introduced in #10

### DIFF
--- a/gh-actions/common/has-diff/action.yml
+++ b/gh-actions/common/has-diff/action.yml
@@ -50,8 +50,8 @@ runs:
 
         # Additional arguments to ignore.
         if [ -n "${{ inputs.regexp-to-ignore }}" ]; then
-          regexp-to-ignore='${{ inputs.regexp-to-ignore }}'
-          diffCmd="${diffCmd} --ignore-matching-lines ${regexp-to-ignore}"
+          regexp_to_ignore='${{ inputs.regexp-to-ignore }}'
+          diffCmd="${diffCmd} --ignore-matching-lines ${regexp_to_ignore}"
         fi
 
         DIFF=$(git difftool -y -x "${diffCmd}" .)


didrocks (dismissed):
I guess you tested it on Windows first and the previous iteration worked there :)

If we are going to do substantiel continous changes to CI, I see 2 ways of avoiding this:
- either have workflow GH action tests:
  - pro: this is a good documentation for the action
  - con: consequent investiment.
- really define a process pointing at least 3 projects (one pure linux, one windows, mono and multi repo) to validate the changes.

EduardGomezEscandell:
> I guess you tested it on Windows first and the previous iteration worked there :)

Not really, I ran it on both Windows and Linux and the worklows passed. I honestly am not sure why it failed now, as this error should have surfaced before.

I think I possibly reverted the wrong thing before merging.

gh-actions/go/code-sanity/action.yaml
```diff
@@ -67,9 +67,13 @@ runs:
           name=$1
           path=$2
           version="$(
-            go mod edit --json | jq -r ".Require[] | select(.Path==\"${path}\") | .Version" || echo 'latest'
+            go mod edit --json | jq -r ".Require[] | select(.Path==\"${path}\") | .Version" || true
```
didrocks, line 70: any reason to use || true if it always exit with 0 for a valid input script?

EduardGomezEscandell: That I forgot.

gh-actions/go/code-sanity/action.yaml
```diff
@@ -67,9 +67,13 @@ runs:
           name=$1
           path=$2
           version="$(
-            go mod edit --json | jq -r ".Require[] | select(.Path==\"${path}\") | .Version" || echo 'latest'
+            go mod edit --json | jq -r ".Require[] | select(.Path==\"${path}\") | .Version"
           )"
-          echo "${name}=${version}" >> $GITHUB_OUTPUT
+          if [ -z "${version}" ]; then
```
didrocks, line 72: I think should that would be more readable btw:
```sh
version="latest"
mod_version="$(
            go mod edit --json | jq -r ".Require[] | select(.Path==\"${path}\") | .Version"
)"
if [ -n "${mod_version}" ]; then
   version="${mod_version}"
fi
echo "${name}=${version}" >> $GITHUB_OUTPUT
```

with:
- default
- If override -> reassign
- action

EduardGomezEscandell: Sounds good

didrocks (approved):
Looking good now. Thanks for the fixes! :)

PR #13, opened 2023-10-20
feat: Provide option to provide pro-token to setup-grpc action

EduardGomezEscandell:
The `arduino/setup-protoc` action uses GitHub API rather than regular workflows. The GitHub API, however, is limited for non-registered users.

This PR allows the caller to specify their GitHub token so that these rates limits are increased.

### DIFF
--- a/gh-actions/go/code-sanity/action.yaml
+++ b/gh-actions/go/code-sanity/action.yaml
@@ -16,6 +16,11 @@ inputs:
   generate-diff-paths-to-ignore:
     description: Files and paths to ignore when checking if generated files changed. This is passed to git update-index.
     default: po/* doc/*.md README.md
+  github-token:
+    description: |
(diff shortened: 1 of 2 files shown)


CarlosNihelton:
How much power are we transfering in sharing this token?

EduardGomezEscandell:
That is up to the caller. By default, it is total and complete control, but that can be changed by setting up permissions: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

I'll experiment with this branch to see what the smallest set of permissions is necessary.

EduardGomezEscandell:
Seems like the only necessary permission is `contents:read`.

EduardGomezEscandell:
Closing because we'd prefer some solution that does not require passing the GITHUB_TOKEN. We'll probably end up finding an alternative to the arduino action.

PR #18, opened 2023-11-28
fix(build-debian): Update apt before installing

EduardGomezEscandell:
Simple enough bugfix. Suprisingly this had not cause any issues until now.


jibel (approved):
obviously. Thanks!

PR #20, opened 2024-01-09
deps: Update actions/upload-artifact from v3 to v4

EduardGomezEscandell:
Very sneaky line in the docs of download-artifact@v4:

> Downloading artifacts that were created from
  action/upload-artifact@v3 and below are not supported.

https://github.com/actions/download-artifact/tree/v4/?tab=readme-ov-file#breaking-changes


didrocks (approved):
Oh right, it’s only us then!

PR #21, opened 2024-01-12
feat: use go version

matthew-hagemann:
Use go version to detect which go toolchain the project is depending on, now that go version does this analyze since go1.21.0.


didrocks (approved):
Excellent!

PR #22, opened 2024-02-06
feat(build-deb): Upload source package and run lintian

EduardGomezEscandell:
Being tested in https://github.com/canonical/ubuntu-pro-for-wsl/pull/546

---

UDENG-2235

### DIFF
--- a/gh-actions/common/build-debian/action.yml
+++ b/gh-actions/common/build-debian/action.yml
@@ -55,10 +55,20 @@ runs:
         DEBIAN_FRONTEND=noninteractive sudo apt install -y devscripts
         echo "::endgroup::"
 
-        echo "::group::Append commit SHA to local version"
+        echo "::group::Create local version with commit and docker container"
         cd '${{ inputs.source-dir }}'
-        sanitized_docker=$( echo "${{ inputs.docker-image }}" | sed 's/://' )
-        debchange --local "+${sanitized_docker}+${{ github.sha }}" "Github build. Job id: ${{ github.run_id }}. Attempt: ${{ github.run_number }}."
+
+        # Sanitize the docker name so that it stick to debian policy
+        # https://www.debian.org/doc/debian-policy/ch-controlfields.html#version
+        sanitized_docker=$( echo "${{ inputs.docker-image }}" | sed -r 's/[^a-zA-Z0-9.+~]+/+/g' )
+
+        # Short commit to avoid "package-has-long-file-name"
+        commit=$(echo ${{ github.sha }} | cut -c1-8)
+
+        export DEBFULLNAME="GitHub actions runner"
+        export DEBEMAIL="noreply@github.com"
+        debchange --local "~${sanitized_docker}+${commit}" "Github build. Run id: ${{ github.run_id }}. Run number: ${{ github.run_number }}."
+        
         echo "::endgroup::"
 
         echo "::group::Parsing name and version"
--- a/gh-actions/common/build-debian/action.yml
+++ b/gh-actions/common/build-debian/action.yml
@@ -106,7 +106,12 @@ runs:
         artifacts-dir: ${{ env.BUILD_OUTPUT_DIR }}
         source-dir: ${{ env.BUILD_INPUT_DIR }}
         docker-image: ${{ inputs.docker-image }}
-    - name: Upload artifacts
+    - name: Copy source package to output dir
+      shell: bash
+      run: |
+        set -eu
+        cp -r ${{ env.SOURCE_OUTPUT_DIR }}/* ${{ env.BUILD_OUTPUT_DIR }}
+    - name: Upload artefacts
       uses: actions/upload-artifact@v4
       with:
         name: ${{ env.PKG_NAME }}_${{ env.PKG_VERSION }}
--- a/gh-actions/common/build-debian/action.yml
+++ b/gh-actions/common/build-debian/action.yml
@@ -12,6 +12,10 @@ inputs:
   token:
     required: false
     description: If provided, used for git authentication in the source build
+  lintian:
+    required: false
+    description: Arguments to pass to lintian, if any. Set to false to skip the lintian chek.
+    default: ''
 
 
 # The process:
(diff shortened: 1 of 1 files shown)


EduardGomezEscandell:
Example failure: https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/7817501950/job/21325536015?pr=546

gh-actions/common/build-debian/action.yml
```diff
@@ -51,10 +55,20 @@ runs:
...
 
-        echo "::group::Append commit SHA to local version"
+        echo "::group::Create local version with commit and docker container"
         cd '${{ inputs.source-dir }}'
-        sanitized_docker=$( echo "${{ inputs.docker-image }}" | sed 's/://' )
-        debchange --local "+${sanitized_docker}+${{ github.sha }}" "Github build. Job id: ${{ github.run_id }}. Attempt: ${{ github.run_number }}."
+
+        # Sanitize the docker name so that it stick to debian policy
+        # https://www.debian.org/doc/debian-policy/ch-controlfields.html#version
+        sanitized_docker=$( echo "${{ inputs.docker-image }}" | sed -r 's/[^a-zA-Z0-9.+~]+/+/g' )
```
3v1n0, line 63: Ideally I'd use `lsb_release -r -s`, but it has to run inside the target docker so that we have an incremental number (while docker names can be static or even cyclic).

gh-actions/common/build-debian/action.yml
```diff
@@ -51,10 +55,20 @@ runs:
...
+
+        # Short commit to avoid "package-has-long-file-name"
+        commit=$(echo ${{ github.sha }} | cut -c1-8)
```
3v1n0, line 66: Why don't we use git ourselves here?

So we can ahve something fancier by using `echo git+$(git rev-list --count HEAD).$(git rev-parse --short HEAD)` (or similar), and we have also an incremental number together with the SHA.

gh-actions/common/build-debian/action.yml
```diff
@@ -51,10 +55,20 @@ runs:
...
+
+        export DEBFULLNAME="GitHub actions runner"
+        export DEBEMAIL="noreply@github.com"
```
3v1n0, line 69: Or maybe latest commit author?

```suggestion
        export DEBFULLNAME="$(git log -1 --format='%an' HEAD) - GH Action"
        export DEBEMAIL="$(git log -1 --format='%ae' HEAD)"
```

Plus an indication that is GH though, to make clear both things.

gh-actions/common/build-debian/action.yml
```diff
@@ -12,6 +12,10 @@ inputs:
   token:
     required: false
     description: If provided, used for git authentication in the source build
+  lintian:
+    required: false
+    description: Arguments to pass to lintian, if any. Set to false to skip the lintian chek.
+    default: ''
```
3v1n0, line 18: This should be `true` IMHO (to make it clearer at the reader as I see that by default it runs).

gh-actions/common/build-debian/action.yml
```diff
@@ -79,6 +93,11 @@ runs:
           if [ -n "${GITHUB_TOKEN}" ]; then
             git config --system url."https://api:${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
           fi
+    - name: Run lintian
+      if: inputs.lintian != 'false'
+      shell: bash
+      working-directory: ${{ env.SOURCE_OUTPUT_DIR }}
+      run: lintian ${{ inputs.lintian }} -- *.changes
```
3v1n0, line 100:
```suggestion
      run: lintian --pedantic --fail-on error ${{ inputs.lintian }} -- *.changes
```

So we still see what's not exactly _clean_. Plus, IMHO, it would be nice to have an option to control `--fail-on`, since ideally good packages should set it to `warning`

3v1n0 (commented):
Thanks for starting this here, I had few other ideas that I was drafting in https://github.com/ubuntu/authd/pull/130, but this seems a good progress.

gh-actions/common/build-debian/action.yml
3v1n0: Oh, I sorry ignore me on the `--fail-on` thingy... It's indeed possible to override the value.

Although, maybe this option is even too powerful, since IMHO any suppression should be in sources.

gh-actions/common/build-debian/action.yml
```diff
@@ -106,7 +106,12 @@ runs:
         artifacts-dir: ${{ env.BUILD_OUTPUT_DIR }}
         source-dir: ${{ env.BUILD_INPUT_DIR }}
         docker-image: ${{ inputs.docker-image }}
-    - name: Upload artifacts
+    - name: Copy source package to output dir
+      shell: bash
+      run: |
+        set -eu
+        cp -r ${{ env.SOURCE_OUTPUT_DIR }}/* ${{ env.BUILD_OUTPUT_DIR }}
```
3v1n0, line 132: IMHO, we should do this phase just after building the source, and ensure that this happens also if binary build fails: This can help a lot to debug situations failure situations, because one can download the sources locally and build them with `sbuild` (or similar tools) to fix the inconsistencies.

Also in commit message, we mostly spell it as "art_i_fact" even if both words seems to be correct.

gh-actions/common/build-debian/action.yml
```diff
@@ -106,7 +106,12 @@ runs:
...
+    - name: Upload artefacts
```
3v1n0, line 133: -> `artifacts`

gh-actions/common/build-debian/action.yml
```diff
@@ -79,6 +83,11 @@ runs:
           if [ -n "${GITHUB_TOKEN}" ]; then
             git config --system url."https://api:${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
           fi
+    - name: Run lintian
```
3v1n0, line 96: Ideally this should run in the same distro in which we build the package... So I'm wondering if we should fork the jtdor/build-deb-action@v1 action to support post-build-hooks or just a run-lintian option.

EduardGomezEscandell:
@3v1n0 I'm a bit busy at the moment, feel free to take over this PR if you can and want. I think I'll be able to work on it in a couple of weeks otherwise.

3v1n0:
> @3v1n0 I'm a bit busy at the moment, feel free to take over this PR if you can and want. I think I'll be able to work on it in a couple of weeks otherwise.

Yeah, nw... I've took something of this in #28 but we can revisit it. Also I didn't add the lintian thing for now as that can probably be split as another job so that could potentially be triggered in parallel with autopkgtests for example

PR #23, opened 2024-02-16
fix(gh-actions): update documentation path in Go code sanity check

GabrielNagy:
Now that we migrated to Read The Docs for project documentation, the directory name changed from `doc` to `docs`, and Markdown files are no longer guaranteed to be at the root of the directory.

Update the default paths for the Go sanity check to reflect this.


didrocks (commented):
Good catch! Can you check other projects using this, like the provisioning one before we go on accepting this?

didrocks (approved):
Excellent, thanks for the change!

PR #24, opened 2024-02-20
feat: add code sanity action for Rust

GabrielNagy:
Similar to our Go action, add one for Rust that integrates the steps from [authd](https://github.com/ubuntu/authd/blob/e6f2ac656fb8e6069bbd4691090acfff2a07d727/.github/workflows/qa.yaml#L70-L89), adding an additional cargo-audit step.

Action in action: https://github.com/GabrielNagy/authd/actions/runs/7975090675

Fixes UDENG-2233

### DIFF
--- a/gh-actions/rust/code-sanity/action.yaml
+++ b/gh-actions/rust/code-sanity/action.yaml
@@ -1,0 +1,68 @@
+name: Rust code sanity check
+description: Checks code against our desktop Rust code quality and process standards.
+
+inputs:
+  token:
+    required: true
+    description: GitHub token required to authenticate certain Rust steps (used to make API calls)
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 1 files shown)


gh-actions/rust/code-sanity/action.yaml
```diff
@@ -0,0 +1,68 @@
+name: Rust code sanity check
+description: Checks code against our desktop Rust code quality and process standards.
+
+inputs:
+  token:
+    required: true
+    description: Token required to authenticate certain Rust steps
```
didrocks, line 7: I think we should be more explicit what this token is, is it the github token? (spoiler alert: I know it’s that one, but without looking at our previous integration version in authd, I would not :p)

didrocks (changes requested):
One documentation related nitpick, otherwise, excellent! Thanks for testing the workflow in authd too!

Nice to see we will have a common standard on Rust quality checks too.

PR #25, opened 2024-02-23
generate: Use go mod edit to install properly versioned tools

3v1n0:
Parse the tools using jq so that we can easily get the tool path and version and install the one explicitly requested

### DIFF
--- a/gh-actions/go/generate/action.yaml
+++ b/gh-actions/go/generate/action.yaml
@@ -24,6 +24,22 @@ outputs:
 runs:
   using: "composite"
   steps:
+    - name: Set up jq
+      shell: bash
+      env:
+        DEBIAN_FRONTEND: noninteractive
+      run: |
+        echo "::group::Download jq"
+        set -eu
+        if [ "${{runner.os}}" = "Windows" ]; then
+          winget.exe install jqlang.jq --accept-source-agreements --accept-package-agreements --silent --verbose || true
+        else
+          SUDO=$(command -v sudo || true)
+          $SUDO apt update
+          $SUDO apt install jq
+        fi
+        echo "::endgroup::"
+        jq --version
     - name: Install tools and dependencies
       id: proto-deps
       working-directory: ${{ inputs.tools-directory }}
@@ -31,7 +47,8 @@ runs:
         echo "::group::Install tools and dependencies"
         set -eu
 
-        tools=$(grep -o '_ ".*"' *.go | cut -d '"' -f 2)
+        tools=$(go mod edit --json | \
+          jq -r '.Require[] | select(.Indirect!=true) | [.Path,.Version] | join("@")')
 
         needsProtoc=false
         for tool in ${tools}; do


gh-actions/go/generate/action.yaml
```diff
@@ -24,14 +24,28 @@ outputs:
 runs:
   using: "composite"
   steps:
+    - name: Set up jq
+      shell: bash
+      run: |
+        echo "::group::Download jq"
```
didrocks, line 30: Please set -eu on the script, like all others, to ensure we stop early if we can’t update/install jq.

didrocks (changes requested):
Thanks for the change, this is indeed an improvement! I have comment before approving it.

gh-actions/go/generate/action.yaml
3v1n0: Oh, weird, I was guess that it was enabled by default (iirc it is in gitlab scripts...).

gh-actions/go/generate/action.yaml
```diff
@@ -24,14 +24,29 @@ outputs:
 runs:
   using: "composite"
   steps:
+    - name: Set up jq
+      shell: bash
+      run: |
+        set -eu
```
didrocks, line 30: Can you put it after the first `echo` (again as in the other steps in this action). That way, we step name will be the echo line and not `set -eu`

didrocks (changes requested):

gh-actions/go/generate/action.yaml
3v1n0: ok, done I wanted to protect also a potential error on `echo` :joy:

didrocks: If echo starts to return 1, I think the script is the least of the concern on this machine! :p

didrocks (approved):
Thanks! Do you have rights to merge it or should I?

3v1n0:
Oh, I think I can once I signed the commit, I did a further tuning though: https://github.com/canonical/desktop-engineering/compare/554197c1ad683d327b381924bede9d38ed8e671e..a118a2ecf9864afc08832b7ebf353303b103b53f

didrocks (changes requested):
I remove then my approval (please don't push new code after the initial review, let's keep the scope focused).

With your latest change, you are not following what is done in other steps in the same file. For the sake of consistency, they should all be changed or none, but please don’t introduce descrepancies there (and so, this is a separate PR with a global env then).

3v1n0:
> With your latest change, you are not following what is done in other steps in the same file.

That's the thing... I did this (and in the same commit) because such file has _not_ anything like that, reason why IMHO it made sense because there was no other code style to follow. Thus it was better doing it in the proper way.

didrocks (approved):
Ack, so can we expect a followup commit to align the other action with this env variable settting?

didrocks:
if you have the merge rights, please change your commit message before merging to use `ci(go): `

3v1n0:
> Ack, so can we expect a followup commit to align the other action with this env variable settting?

YEah... I was thinking at it for a while...

PR #26, opened 2024-02-23
ci(go): Cleanup some job steps

3v1n0:
Cleanup the code and expose variables when required, as discussed in #25


didrocks (commented):
I guess after #25, please provide a link of a successful github action run with those changes.

3v1n0:
> I guess after #25, please provide a link of a successful github action run with those changes.

I feel it would be better and more maintainable to add some example jobs to this repo to test the actions on each MR, so I'd rather prefer doing that once I've time for.

didrocks:
I’m not going to merge something that is impacting a lot of project without having a link though :p Even if the changes looks trivial, as the previous PR failure, I don’t think we should trust not spotting any typos.

PR #27, opened 2024-02-23
fix(gh-actions): Use the full command path to go install a tool 

3v1n0:
Uff, sadly #25 lead to an unexpected issue when working with some tools, so changed the logic a bit still preserving the `go mod` usage but also fetching the full path through unix tools.

### DIFF
--- a/gh-actions/go/generate/action.yaml
+++ b/gh-actions/go/generate/action.yaml
@@ -24,6 +24,29 @@ outputs:
 runs:
   using: "composite"
   steps:
+    - name: Set up jq
+      shell: bash
+      env:
+        DEBIAN_FRONTEND: noninteractive
+      run: |
+        echo "::group::Ensure jq is installed"
+        set -eu
+
+        if command -v jq &> /dev/null; then
+          echo "::endgroup::"
+          exit 0
+        fi
+
+        set -eu
+        if [ "${{runner.os}}" = "Windows" ]; then
+          winget.exe install jqlang.jq --accept-source-agreements --accept-package-agreements --silent --verbose || true
+        else
+          SUDO=$(command -v sudo || true)
+          $SUDO apt update
+          $SUDO apt install jq
+        fi
+        echo "::endgroup::"
+        jq --version
     - name: Install tools and dependencies
       id: proto-deps
       working-directory: ${{ inputs.tools-directory }}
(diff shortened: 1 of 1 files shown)


3v1n0:
Can check the result is now fine in https://github.com/3v1n0/authd/actions/runs/8026458635/job/21928966323 (not sure if you can see the logs though (so [here](https://pastebin.ubuntu.com/p/7m8CgtH5rj/) it is).

didrocks:
I’m rather reverting to unblock the other jobs. I have some comments on that one and don’t want to block further.

Please, be mindful of any change in that project and always run some Github Actions against your proposed version first. I thought you would have done that (you are the first to break it, congrats :p)

gh-actions/go/generate/action.yaml
```diff
@@ -29,6 +29,10 @@ runs:
       env:
         DEBIAN_FRONTEND: noninteractive
       run: |
+        if command -v jq &> /dev/null; then
```
didrocks, line 32: The steps will be named "if command…".

I suggest changing the "Download jq" echo to "Ensure jq is installed", and put that at the first line.

Also, then, your commands will be after `set -eu`

gh-actions/go/generate/action.yaml
```diff
@@ -50,12 +54,17 @@ runs:
...
-        needsProtoc=false
         for tool in ${tools}; do
+          IFS='@' read -r -a pair <<< "${tool}"
+          echo "${pair[0]}"
           if [[ "${tool}" == *protoc* ]]; then
             echo "needs-protoc=true" >> $GITHUB_OUTPUT
           fi
-          go install ${tool}
+
+          cmd=$(grep -F "${pair[0]}" ./*.go | sed -nr 's/.*_\s+"([^"]+)".*/\1/p' | head -1)
```
didrocks, line 64: Add a command to gives examples against what this regexp against and how.

didrocks (changes requested):
Please incorporate back the reverted changes.

However, seeing the new regexp, I really feel this json version is now more complex than the one we had.
Also, you force parsing the version, and `go install` that specific version, while the previous code was using go itself, which will detect the version as it’s in the correct directory. What are you trying to fix with parsing the json file and added more lines to maintain?

gh-actions/go/generate/action.yaml
3v1n0: The point of it is:
 1. Grep with regex disabled for the line containing the package URI
 2. Strip from a line such as `     _     "URI" ` the `URI`
 3. Ensure we have no duplicates.

By add a command what you mean? Add a comment with an example?

didrocks: Sorry, I meant "comment"

didrocks (changes requested):
I don’t understand why you are continuing pushing code on that one after our MM discussion.

As said there:
- there is nothing to fix, the current approach was strong and resilient
- trying to use the json parsing slow down the job (we need to detect, install jq, support this on both Windows and Linux with different mean of installing it), and then brings the issues that only modules are shown
- we now need then, instead of an easy grep, do a `read -r`, split the version and then, having a `grep -F | sed | head -n1`.
- it doesn’t let `go` installing from the directory, which will autodetect the version it resolves to. Here, you are assuming the version your extracted it the exact one it will use, but as told on MM, `go` will decide this in a reproducible way based on the version in `.mod`, but the MVS algorithm may pick a different one based on vulnerability checking, toolchain compatibility and potential dependencies interaction. Enforcing a version is then just wrong. Just `cd` to the directory, `go install`, and we are done! (and this is what the existing code in main does)

Sorry, but this is adding complexity for complexity and not solving anything. We saw that it already broke and there is still no link of a successful run, I would really suggest closing that one.

3v1n0:
Yeah, that's fine... It's one of those cases were apparently simple cleanups are more expensive than expected, so we can continue this way I think.

> We saw that it already broke and there is still no link of a successful run, I would really suggest closing that one.

On that though... It was there since the beginning: https://github.com/canonical/desktop-engineering/pull/27#issuecomment-1962156953 :)

didrocks:
> On that though... It was there since the beginning: https://github.com/canonical/desktop-engineering/pull/27#issuecomment-1962156953 :)

right, but you did push more fixes and rerequest a review without a new run linked in the description :)

PR #35, opened 2024-08-29
maint: update golangci-lint action

didrocks:
This update allows to remove an hack that was previously necessary. Also, we need v6 now for GH to accepts the annotations format.

UDENG-4223


jibel (approved):
+1 thanks!

PR #37, opened 2024-11-12
feat: including pulse creation logic from jira-tools.py in the main jira script

sminez:
Rename `jira.py` to `jira-tools.py` and expose its functionality through the main `jira` script.


sminez:
That makes sense, yeah. I was just wanting to avoid nested subcommands with argparse on the python side as it's a faff but we can sort that out pretty easily. I've not got time to work on these scripts any more this week I don't think though so I can pick it up next week or leave it for someone else to pick up if they fancy it :wink:

didrocks (approved):
Again, I don’t want to block on this. Let’s see once we tackle that and merging this.

PR #39, opened 2024-11-12
feat: allow setting new cards descriptions from the cmdline

seb128:
A non interactive mode makes it easier to use in a script


sminez (approved):
:+1:

PR #40, opened 2024-11-13
ci(build-deb): Perform lintian checks on sources and binary packages

3v1n0:
UDENG-5415

### DIFF
--- a/gh-actions/common/build-debian/action.yml
+++ b/gh-actions/common/build-debian/action.yml
@@ -59,7 +59,7 @@ runs:
         cd '${{ inputs.source-dir }}'
 
         # Short commit to avoid "package-has-long-file-name"
-        echo VERSION_REF=$(date +'%y%m%d').${{ github.run_number }}.$(echo ${{ github.sha }} | cut -c1-8) >> $GITHUB_ENV
+        echo VERSION_REF=$(date +'%y%m%d')+${{ github.run_number }}+$(echo ${{ github.sha }} | cut -c1-8) >> $GITHUB_ENV
 
         echo DEBFULLNAME="GitHub actions runner" >> $GITHUB_ENV
         echo DEBEMAIL="noreply@github.com" >> $GITHUB_ENV
--- a/gh-actions/common/build-debian/action.yml
+++ b/gh-actions/common/build-debian/action.yml
@@ -299,9 +299,24 @@ runs:
 
           echo "::group::Show binaries information"
           for i in "${{ env.BUILD_OUTPUT_DIR }}"/*.deb; do
-            echo "$(basename "$i")"
-            dpkg --info "$i"
-            dpkg --contents "$i"
+            {
+              pkg_name=$(basename "$i")
+              echo '## Package `'"${pkg_name}"'`'
+              echo
+              echo '```'
+              echo "❯ dpkg --info $pkg_name"
+              dpkg --info "$i" 2>&1
+              echo '```'
+              echo
+              echo "<details>"
+              echo "<summary>Package contents</summary>"
+              echo
+              echo '```'
+              echo "❯ dpkg --contents $pkg_name" 2>&1
+              dpkg --contents "$i"
+              echo '```'
+              echo "<details>"
+            } | tee -a "${GITHUB_STEP_SUMMARY}"
           done
           echo "::endgroup::"
 
--- a/gh-actions/common/build-debian/action.yml
+++ b/gh-actions/common/build-debian/action.yml
@@ -17,6 +17,10 @@ inputs:
     required: false
     # FIXME: this should default to '', but we don't want to break job depending on us for now
     default: 'ca-certificates git'
+  lintian:
+    required: false
+    description: Arguments to pass to lintian, if any. Set to `skip` to skip the lintian check.
+    default: ''
 
 
 # The process:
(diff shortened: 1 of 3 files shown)


3v1n0:
FYI (@CarlosNihelton, @ashuntu) in the wsl pro client with this we only have two minor warnings:

On sources:
```
 W: wsl-pro-service source: superfluous-file-pattern vendor/go.uber.org/atomic/* [debian/copyright:100]
```

On binaries:
```
  W: wsl-pro-service: hardening-no-pie [usr/libexec/wsl-pro-service]
  P: wsl-pro-service: spelling-error-in-copyright Unknwon Unknown
```

But none of these leads to a build failure.

gh-actions/common/build-debian/action.yml
```diff
@@ -17,6 +17,10 @@ inputs:
     required: false
     # FIXME: this should default to '', but we don't want to break job depending on us for now
     default: 'ca-certificates git'
+  lintian:
+    required: false
+    description: Arguments to pass to lintian, if any. Set to false to skip the lintian check.
```
didrocks, line 22: and why not `skip`?

gh-actions/common/build-debian/action.yml
```diff
@@ -160,6 +164,40 @@ runs:
         path: ${{ env.SOURCE_OUTPUT_DIR }}/
         if-no-files-found: error
 
+    - name: Run lintian on sources
```
didrocks, line 167: I would be good to explain why you want to run lintian in a container and not just installing it as part of dev package. (I do see you create an user, but I’m blanking on why)

gh-actions/common/build-debian/action.yml
```diff
@@ -160,6 +164,40 @@ runs:
...
+
+          echo "::group::Create test user"
+          adduser --disabled-password --gecos "" tester
+          chown tester:tester . -R
+          echo "::endgroup::"
+
+          echo "::group::Run lintian"
+          runuser -u tester -- \
+            lintian --pedantic --fail-on error ${{ inputs.lintian }} -- \
+              "${{ env.PKG_NAME }}_${{ env.PKG_VERSION }}_source".changes
```
didrocks, line 198: Do you think we can annotate the Github Action CI build itself with the debug:warnings/error from it? If you want to see how to write something there, you can use something like https://github.com/canonical/desktop-engineering/blob/main/gh-actions/common/print-summary/action.yml#L33. Or you can format special warnings to be printed.

didrocks (changes requested):
2 suggestions and one remark that needs clarification! Otherwise great! :)

gh-actions/common/build-debian/action.yml
3v1n0: Yeah, you're reading my mind on that.... I was in fact actually working on a script to generate that right now :-D

gh-actions/common/build-debian/action.yml
3v1n0: Well, it brings a lot of dependencies in, so I don't want end up in creating false positives by being able to build a source package when the required -dev dependencies are not yet installed.

gh-actions/common/build-debian/action.yml
3v1n0: ah, `skip` that's fine too, I think I was basing this choice on some older PR that was suggested

gh-actions/common/build-debian/action.yml
didrocks: mind adding a comment for it?

3v1n0:
| Result | Pacakge | Tag | Details |
|:--:|:--:|--|--|
| 🟠 |foo-pkg |`absolute-symbolic-link-target-in-source` | |
| 🟠 |foo-pkg |`absolute-symlink-in-top-level-folder` | |
| 🟡 |foo-pkg |`acute-accent-in-manual-page` | |
| 🟡 |foo-pkg |`add-component-copyright` | |
| 🟠 |foo-pkg |`adduser-with-home-var-run` | |
| 🟡 |foo-pkg |`adopted-extended-field` | |
| 🔴 |foo-pkg |`alternates-not-allowed` | |
| 🟠 |foo-pkg |`alternatively-build-depends-on-python-sphinx-and-python3-sphinx` | |
| 🟠 |foo-pkg |`ambiguous-paragraph-in-dep5-copyright` | |
| 🟠 |foo-pkg |`ancient-libtool` | |
| 🟠 |foo-pkg |`ancient-python-version-field` | |
| 🟠 |foo-pkg |`ansi-escape` | |
| 🟡 |foo-pkg |`anticipated-repack-count` | |
| 🔴 |foo-pkg |`apache2-configuration-files-need-conf-suffix` | |
| 🟠 |foo-pkg |`apache2-deprecated-auth-config` | |
| 🔴 |foo-pkg |`apache2-module-does-not-depend-on-apache2-api` | |
| ⚪ |foo-pkg |`wiki-copyright-format-uri` | |
| ⚪ |hello |`copyright-refers-to-symlink-license` |usr/share/common-licenses/GPL |
| 🟠 |wsl-pro-service |`hardening-no-pie` |[usr/libexec/wsl-pro-service] |
| ⚪ |wsl-pro-service |`spelling-error-in-copyright` |Unknwon Unknown |

gh-actions/common/build-debian/action.yml
3v1n0: Ok done

didrocks (approved):
Great work! That almost tempt me to get some lintian bugs on my packages to see this beautiful summary! :)

PR #41, opened 2024-11-13
feat: add an option to not assign the new cards

seb128:
the arguments parsing will be used as a base to add other options

### DIFF
--- a/bin/jira
+++ b/bin/jira
@@ -90,15 +90,35 @@ EOF
 
 
 function create_issue_in_current_sprint {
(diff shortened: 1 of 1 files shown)


sminez (approved):

seb128:
I will also add options to set an epic or to add the card to the backlog instead of the current pulse. The current parsing should make it easy enough but I wonder if maybe we should consider using getopts instead?

Also the usage summary needs to be updated/reformatted to describe the options but I will wait on having a consensus on current parsing vs getopts before doing extra changes

PR #45, opened 2024-11-14
feat(project-repo): standardize github settings and commit types

d-loose:
We've recently discussed review process standards and corresponding github settings that help support this process in one of our weekly meetings with @didrocks. This PR contains a few proposed changes and simplifications that came out of this meeting.

I'd be happy to iterate on this a bit further in case someone has concerns or more ideas.

UDENG-5073

### DIFF
--- a/project-repo/github-settings.md
+++ b/project-repo/github-settings.md
@@ -1,0 +1,26 @@
+# Recommended GitHub repository settings
+
+In order to encourage all contributors to follow the recommended [review process](review-process.md), the repository settings listed in the following sections should be applied.
+
+## Pull requests
+
+The settings listed below should be enabled/disabled as indicated. All other settings can be chosen based on project-specific needs.
+
+- [x] Allow merge commits (default commit message: "Pull request title and description")
+- [ ] Allow squash merging
+- [ ] Allow rebase merging
+
+- [x] Automatically delete head branches
+
+
+## Branch protection rules
+
+The `main` branch (and any other public development branches from which releases are created) should be protected with by a corresponding set of rules.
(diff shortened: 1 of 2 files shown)


sminez (approved):

didrocks (commented):
Thanks for formalising this!

My comments are around discussing on the "finale roundtrip" on which I’m not fully for and against it, as there are advantages and drawbacks in both approaches.

On the settings (for branch protection): we generally have short circuit fallback to allow some people to force push if needed to main (zomg). This is generally to unblock situations and be a little bit like a disaster recovery, I don’t know if you want to include this here?

As an aside, we should also standardize around the "Code Security" scanning for both active scanning and dependabot settings.

d-loose:
This has lost a bit of momentum due to other priorities, but if we want to continue thinking about establishing a standard process, we'd probably need to agree on a few constraints first (judging by the comments so far):
- Who is supposed to follow this process / impose it in their projects? (individual squads, the entire desktop team)
- How strict do we want the definition to be? @rcaballeromx suggested we should rather have a strict definition and a lenient imposition than the other way around. (Also @sminez keeps saying that "everyone SHOULD read [RFC2119](https://www.ietf.org/rfc/rfc2119.txt)")
- Do we want to define more than one process? What @3v1n0 describes is incompatible with what's currently defined here.

didrocks:
In an ideal world, that would be a process shared company-wise. As it’s not realistic to expect that from the get go, I think the only possible option is the other way around: onboarding squads on them, one after another.

We are quite strict at imposing that in the enterprise desktop team and WSL and following it (with the last rebase trusting the submitter though, which only created one regression due to the rebase itself in the past 3 years, which I think is more than a good score with the flexibility this provides).

PR #46, opened 2024-11-18
fix: correcting key for setting story points

sminez:
When I made story points for tickets optional I failed to correctly nest the key under `fields` when setting the custom field outside of the in initial payload creation.


seb128 (approved):
Thanks, makes sense

PR #47, opened 2024-11-18
gotestfmt: Group output of successful tests

adombeck:


3v1n0:
Cool, it would be probably nice to add a CI check for using `gotestfmt` inside this repo so that we can ensure that the changes are working without depending on other branches (outside this repo), but the change looks fine to me

3v1n0 (approved):

PR #48, opened 2024-11-22
Allow to list epics restricted to specific components

seb128:


seb128:
The components argument is optional, you can also include several components e.g

`$ ./bin/jira list-epics --backlog=UDENG --components=Integration,Maintenance`

sminez (approved):

PR #49, opened 2024-11-22
fix: rename backlog argument to project

seb128:
the use of backlog in that context is confusing


seb128:
It might be my JIRA knowledge being off but for me 'backlog' is the cards not in a pulse, I find it confusing as an argument for `list-epics` (and it seems used as a value for $project then in code)?

sminez (approved):
This one's force of habit on my part. For whatever reason at my last job everyone referred to Jira projects as backlogs 🤷

PR #54, opened 2024-12-14
feat(gh-actions/build-debian): Support running a workflow script at build-source phase

3v1n0:
When building some packages we may need to prepare the build environment before preparing the source package, for example in some cases we may need to enable git access (to compute the package version or similar) or we may need to install tools that are used during debian source building that are not provided in the archive (such as rust crates, hello cargo-vendor-filterer!!).

So instead of polluting this generic action with package-specific requirements, let's just make it support running a script that can do more advanced configurations in the docker instance that is used only during the source-preparation phase.

Not covering the migration yet (since I'd like to provide some more tools for it), but I think that we should drop the specific-git code from this action too, and move it where it's relevant (in the ubuntu-pro-for-wsl repo as action input script instead).

This is required in order to be able to do a `cargo install --root=/usr cargo-vendor-filterer` to build the source package of rust programs as required by authd in https://github.com/ubuntu/authd/pull/130

UDENG-2343

### DIFF
--- a/gh-actions/common/build-debian/action.yml
+++ b/gh-actions/common/build-debian/action.yml
@@ -17,6 +17,13 @@ inputs:
     required: false
     # FIXME: this should default to '', but we don't want to break job depending on us for now
     default: 'ca-certificates git'
+  extra-source-build-script:
+    description: |
+      A script to run to prepare the source build machine.
+      This happens after the dependencies have been installed, but before
+      running `dpkg-buildpackage -S`.
+    required: false
+    default: ''
   lintian:
     required: false
     description: Arguments to pass to lintian, if any. Set to `skip` to skip the lintian check.
@@ -147,6 +154,15 @@ runs:
           GITHUB_TOKEN="${{ inputs.token }}"
           if [ -n "${GITHUB_TOKEN}" ]; then
             git config --system url."https://api:${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
+          fi
+
+          if [ -n "${{ inputs.extra-source-build-script != '' && 'true' || '' }}" ]; then
+            echo "::group::Run source build script"
+            (
+              set -eux
+              ${{ inputs.extra-source-build-script }}
+            )
+            echo "::endgroup::"
           fi
 
           echo "::group::Build debian source package"
--- a/gh-actions/common/build-debian/action.yml
+++ b/gh-actions/common/build-debian/action.yml
@@ -16,7 +16,7 @@ inputs:
     description: A list of extra build dependencies required during source build.
     required: false
     # FIXME: this should default to '', but we don't want to break job depending on us for now
-    default: 'ca-certificates git'
+    default: 'ca-certificates'
   extra-source-build-script:
     description: |
       A script to run to prepare the source build machine.
@@ -140,9 +140,17 @@ runs:
 
           echo "::group::Install build dependencies"
           apt build-dep .
+
+          GITHUB_TOKEN="${{ inputs.token }}"
+
           if [ -n "${{ inputs.extra-source-build-deps }}" ]; then
             # Install extra packages for build-deps, to allow downloading vendored sources
             deps=(${{ inputs.extra-source-build-deps }})
+
+            if [ -n "${GITHUB_TOKEN}" ]; then
+              deps+=(git)
+            fi
+
             apt install ${deps[@]}
           fi
           echo "::endgroup::"
@@ -151,7 +159,6 @@ runs:
             git config --system --add safe.directory "${{ github.workspace }}"
           fi
 
-          GITHUB_TOKEN="${{ inputs.token }}"
           if [ -n "${GITHUB_TOKEN}" ]; then
             git config --system url."https://api:${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
           fi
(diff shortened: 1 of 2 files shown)


gh-actions/common/build-debian/action.yml
```diff
@@ -17,6 +17,13 @@ inputs:
     required: false
     # FIXME: this should default to '', but we don't want to break job depending on us for now
     default: 'ca-certificates git'
+  extra-source-build-script:
+    description: |
+      A script with commands to run to prepare the source build machine.
+      This happens after that the dependencies have been installed, but before
```
adombeck, line 23:
```suggestion
      This happens after the dependencies have been installed, but before
```

gh-actions/common/build-debian/action.yml
```diff
@@ -17,6 +17,13 @@ inputs:
     required: false
     # FIXME: this should default to '', but we don't want to break job depending on us for now
     default: 'ca-certificates git'
+  extra-source-build-script:
+    description: |
+      A script with commands to run to prepare the source build machine.
```
adombeck, line 22:
```suggestion
      A script to run to prepare the source build machine.
```

gh-actions/common/build-debian/action.yml
```diff
@@ -140,6 +147,9 @@ runs:
           fi
           echo "::endgroup::"
 
+          # FIXME: This part should moved to ${{ inputs.extra-source-build-script }}
```
adombeck, line 150:
```suggestion
          # FIXME: This part should be moved to ${{ inputs.extra-source-build-script }}
```

gh-actions/common/build-debian/action.yml
```diff
@@ -140,6 +147,9 @@ runs:
...
+          # Since it's very project-specific, as most of projects don't care about
+          # Having git at all when building the package source.
```
adombeck, line 152:
```suggestion
          # Since it's very project-specific, as most projects don't care about
          # having git at all when building the package source.
```

gh-actions/common/build-debian/action.yml
```diff
@@ -140,6 +147,9 @@ runs:
...
           if command -v git &> /dev/null; then
             git config --system --add safe.directory "${{ github.workspace }}"
           fi
```
adombeck, line 155: This actually seems like a quite generic thing to me (ensuring that git can be executed in the workspace directory even when it has a different owner, when git is installed)

gh-actions/common/build-debian/action.yml
```diff
@@ -148,6 +158,16 @@ runs:
           if [ -n "${GITHUB_TOKEN}" ]; then
             git config --system url."https://api:${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
           fi
+          # END of FIXME
+
+          if [ -n "${{ inputs.extra-source-build-script != '' && 'true' || '' }}" ]; then
```
adombeck, line 163: Huh, I didn't know that GitHub allows this kind of expression evaluation, but it seems like it does: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#operators

TIL!

adombeck (approved):

adombeck:
Alternatively to specifying a new input, we could also always check if a script `/usr/bin/pre-build-hook` (or something like that) exists and execute it if it does.

gh-actions/common/build-debian/action.yml
3v1n0: Yeah, it's useful here, because otherwise I'd have to care about bash escaping, that wouldn't really work well unless using extra magic... Reason why I preferred this strategy :)

3v1n0:
> Alternatively to specifying a new input, we could also always check if a script /usr/bin/pre-build-hook (or something like that) exists and execute it if it does.

The problem of that is that it would imply saving something in a place (the docker instance) that the workflow has no access to, so that's something we can't do easily. Indeed we could allow the input to be a file instead that can be copied there though.

But I feel this syntax still allows to keep everything in the same place (the calling workflow file) easily. Indeed we could support injecting another file in future though.

gh-actions/common/build-debian/action.yml
3v1n0: Well this is maybe true... But the rest of things (like the github token looks quite per-repo specific IMHO).

adombeck: > But the rest of things (like the github token looks quite per-repo specific IMHO)

does it? it configures git to use the `GITHUB_TOKEN` in its URL. that doesn't look repo-specific to me.

adombeck:
> The problem of that is that it would imply saving something in a place (the docker instance) that the workflow has no access to, so that's something we can't do easily. 

Ah, so intend to provide the file via the git repo, not via the Docker image? Then we could also check `${{ github.workspace }}/bin/pre-build-hook` or something.

> But I feel this syntax still allows to keep everything in the same place (the calling workflow file) easily. Indeed we could support injecting another file in future though.

Works for me.

3v1n0:
> Ah, so you intend to provide the script via the git repo, not via the Docker image? Then we could also check ${{ github.workspace }}/bin/pre-build-hook

yeah... Ideally such path should be provided, but with this way it's still possible to just launch it manually with one line, since the workspace is always already mounted... So I'd keep this way still.

If something more complex is needed, we can reiterate :)

gh-actions/common/build-debian/action.yml
3v1n0: Yeah, that's fine... I've dropped the FIXME already and we do install git when that's defined :)

3v1n0:
This has been tested working in the only customer (so far) so let's merge it. :)

PR #56, opened 2024-12-17
feat(gh-actions): Add an action for checking C format

sergio-costas:
This action and script allows to automatically check the format style in a C/C++ project, using clang-format. It also can be configured as a GIT hook, to also check it automatically before a commit.

### DIFF
--- a/gh-actions/C/format-source.sh
+++ b/gh-actions/C/format-source.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# Configure here all the directories where are source files
+SOURCE_FILES="src/*.[ch] tests/*.[ch]"
+
+if [[ "$1" == "pre-commit" ]]; then
+    echo Checking source style
+    PRE_COMMIT=1
+else
+    PRE_COMMIT=0
+fi
+
+passed=true
+for file in $SOURCE_FILES; do
+    if [ $# -eq 0 ]; then
+        # no parameters? Just apply the changes
+        echo Formating $file
+        clang-format -i $file
+    else
+        # any parameter? check that the formatting is fine
+        clang-format $file > $file.formatted
+        echo Checking $file
+        if [ $PRE_COMMIT -eq 0 ]; then
+            diff $file $file.formatted
+        else
+            diff $file $file.formatted > /dev/null
+        fi
+        if [ $? != 0 ]; then
+            passed=false
+        fi
+        rm $file.formatted
+    fi
+done
+if [ $passed = false ]; then
+    echo Failed to pass clang-format check
+    exit 1
+fi
--- a/gh-actions/C/format.yaml
+++ b/gh-actions/C/format.yaml
@@ -1,0 +1,18 @@
+name: Formatting
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Verify formatting
+        run: |
+          ./format-source.sh check
--- a/gh-actions/C/README.md
+++ b/gh-actions/C/README.md
@@ -1,0 +1,28 @@
+# Check source formatting
+
+This folder contains both a script (`format-source.sh`) and a gh-action
+(`format.yaml`) for checking and fixing the format of the source code,
+both locally during the commit operation and in the CI steps during
+a `git push`.
+
+## Configuring
+
+Just copy the `format.yaml` file to the `.github/workflows/` folder and
+the `format-source.sh` script to the project root folder, and edit the
+script to configure the folders where to search for source code files
+(both .c and .h).
+
+Also, edit the `.git/hooks/pre-commit` script:
+
+    #!/bin/sh
+    ./format-source.sh pre-commit
+
+and give to it execution permissions.
+
+Now, every time a `commit` is done, the source code will be checked and,
+if it doesn't follow the desired format, the commit operation will be
+aborted.
+
+Just executing the `format-source.sh` script from the command line will
+automatically format the source following the style defined in the
+`.clang-format` file (or the `clang-format` defaults if it doesn't exist).


gh-actions/C/format-source.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
3v1n0, line 1:
```suggestion
#!/usr/bin/env bash
```

gh-actions/C/format-source.sh
3v1n0, line 1: As per `shellcheck`:

```
❯ shellcheck /tmp/checker.sh

In /tmp/checker.sh line 17:
        echo Formating $file
                       ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
        echo Formating "$file"


In /tmp/checker.sh line 18:
        clang-format -i $file
                        ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
        clang-format -i "$file"


In /tmp/checker.sh line 21:
        clang-format $file > $file.formatted
                     ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.
                             ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
        clang-format "$file" > "$file".formatted


In /tmp/checker.sh line 22:
        echo Checking $file
                      ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
        echo Checking "$file"


In /tmp/checker.sh line 24:
            diff $file $file.formatted
                 ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.
                       ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
            diff "$file" "$file".formatted


In /tmp/checker.sh line 26:
            diff $file $file.formatted > /dev/null
                 ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.
                       ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
            diff "$file" "$file".formatted > /dev/null


In /tmp/checker.sh line 28:
        if [ $? != 0 ]; then
             ^-- SC2181 (style): Check exit code directly with e.g. 'if ! mycmd;', not indirectly with $?.


In /tmp/checker.sh line 31:
        rm $file.formatted
           ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
        rm "$file".formatted

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
  https://www.shellcheck.net/wiki/SC2181 -- Check exit code directly with e.g...
```

gh-actions/C/format-source.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Configure here all the directories where are source files
+SOURCE_FILES="src/*.[ch] tests/*.[ch]"
```
3v1n0, line 4: These should be configurables as an action `input`, ideally you can just rely on `find \( -name '*.c' -or -name '*.h' \) -exec sh -c 'clang-format --dry-run --Werror "$1" || kill $PPID' -- {} \;`

gh-actions/C/format.yaml
```diff
@@ -0,0 +1,18 @@
+name: Formatting
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
3v1n0, line 7: This should not be defined here

gh-actions/C/format-source.sh
sergio-costas: Good point! Fully agree.

gh-actions/C/format-source.sh
sergio-costas: You are right. Maybe the exit code check could be discussed, but everything else must be applied yes or yes. Thanks!

gh-actions/C/format-source.sh
sergio-costas: Mmm... good point... Give me a moment.

sergio-costas: The point is that the same script must be able to be run from the shell: you want to also run it from pre-commit hook to avoid unnecessary CI tests, and manually to format the code. And the idea is to use exactly the same script in those three cases to ensure consistency (in fact, this idea came when I formatted a commit with what I believed was a correct clang-format configuration, only to discover after the upload that it wasn't). That means that we can't use action inputs. About using find... it can be a good idea. I'll give it a spin.

sergio-costas: Ok, I gave it a spin, and has the problem that it also formats any .c and .h file inside the `build` folder too (the ones generated by meson; for example the code generated by `gdbus-codegen`). That's not really a problem, just an annoyance. Anyway, here is the change. It required some tricks to work, anyway... Give you opinion, please.

gh-actions/C/format.yaml
3v1n0: In the action using it

3v1n0:
I think it would also nice to include some tests for this action, so adding some workflows to that test both scenarios (a formatting succeeded and a failure).

gh-actions/C/format.yaml
sergio-costas: So you mean to remove those lines, and let each team/developer to modify the action and put what it need...?

gh-actions/C/format-source.yaml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Verify formatting
+        run: |
+          ./format-source.sh check
```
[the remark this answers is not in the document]
sergio-costas: Done

gh-actions/C/format.yaml
sergio-costas: I checked the debian one. Done.

gh-actions/C/format-source/format-source.sh
```diff
@@ -0,0 +1,69 @@
...
+}
+
+RESULT=""
+
+if [ -n "${SOURCE_PATHS}" ]; then
+# if SOURCE_PATHS is defined, search only in the specified paths
+    for source_path in ${SOURCE_PATHS}; do
+        RESULT=${RESULT}$(find $source_path -name '*.c' -or -name '*.h' | while read file; do
+            check_source_file "$file"
+        done)
```
nteodosio, line 49: There are a couple of edge cases in which this fails. For example, consider this (outrageous) tree:
```
/tmp/pull56% tree
./
├── a*/
│   └── keyword_table.c
├── aaa/
│   └── keyword_table.c
├── -b/
│   └── keyword_table.c
├── c/
├── x\\/
│   └── keyword_table.c
└── pull56.sh*
```

For `SOURCE_PATHS='-b c x\\ a*'`, this piece of code will produce
```
find: unknown predicate `-b'
check_source_file x\/keyword_table.ccheck_source_file a*/keyword_table.c aaa/keyword_table.c check_source_file aaa/keyword_table.ccheck_source_file aaa/keyword_table.c
```
Note that it gave an error (which could be more confused were the directory named `-name` for instance) and matched the directory `aaa`, which is not in `SOURCE_PATHS`, and it tried `x\` instead of `x\\`.

I'd suggest:
```
for source_path in ${SOURCE_PATHS}; do
    source_path=$(realpath -- "$source_path")
    RESULT=${RESULT}$(find "$source_path" \( -name '*.c' -o -name '*.h' \) -exec check_source_file {} \;)
done
```
It produces
```
check_source_file /tmp/pull56/-b/keyword_table.ccheck_source_file /tmp/pull56/x\\/keyword_table.ccheck_source_file /tmp/pull56/a*/keyword_table.ccheck_source_file /tmp/pull56/aaa/keyword_table.c
```

Of course this still fails for paths containing whitespaces and this is annoyingly nitpicky, but I think the proposed snipped is simpler to follow too, that's why I suggest it. (:

gh-actions/C/format-source/format-source.sh
```diff
@@ -0,0 +1,69 @@
...
+# also, since find and while read... are executed in a subshell, they can't
+# modify any variable outside. That's why we capture the output in $RESULT
+# and compare it with an empty string.
+# Of course, for this to work, it is mandatory to send any visible message to
+# STDERR. That's why the check_source_file() function has so many >&2: those
+# are commands whose output we want in the screen. This trick is needed because
+# there seems to not be an easy way of storing STDERR instead of SDTOUT
+    RESULT=${RESULT}$(find -name '*.c' -or -name '*.h' | while read file; do
+        check_source_file "$file"
+    done)
```
nteodosio, line 63: Can we also simplify to
```
RESULT=${RESULT}$(find \( -name '*.c' -o -name '*.h' \) -exec check_source_file "$file" {} \;)
```
here? Also isn't RESULT guaranteedly empty when this line is reached, so we could drop `${RESULT}` from it?

nteodosio (changes requested):

gh-actions/C/format-source/format-source.sh
sergio-costas: Good point, indeed!!!!

gh-actions/C/format-source/format-source.sh
sergio-costas: Mmm... this doesn't work because `check_source_file` isn't a command, but a function in the .sh file... :-(

gh-actions/C/format-source/format-source.sh
sergio-costas: The same than below: check_source_file is a function, not an external command, so it can't be executed by `find` :-( But the idea of putting the full path is good.

sergio-costas: Changed. Also, will try to fix the blank spaces problem. Thanks!

sergio-costas:
@nteodosio Ok, I did some tests, and there is no problem if there are folders inside with blank spaces. Also, sent a fix for the case where any of the parent folders has blank spaces. The only problem is that it's not possible to add folders with blank spaces to the SOURCE_PATHS environment variable, but I don't think that's a true problem.

gh-actions/C/format-source/format-source.sh
nteodosio: Couldn't we simply move it to a separate file to make it a command and then invoke it with `find -exec /path/to/check_source_file`? Not that this is blocking.

sergio-costas: Mmm... I think that it makes things more complex, because it means two "alien" files in the project instead of one.

nteodosio (approved):

gh-actions/C/format-source/format-source.sh
nteodosio: Alright, I'm approving, sure it is no use blocking on nitpicks especially when our projects won't surely have weirdly named files in it and even if they did this won't cause breakage in the project itself, just in the CI.

PR #57, opened 2024-12-18
cleanup(gh-actions/build-debian): Unify steps common tasks and make apt faster

3v1n0:
Simplify the action and move the common docker steps in a script that is shared by all the steps.

### DIFF
--- a/gh-actions/common/build-debian/prepare-container.sh
+++ b/gh-actions/common/build-debian/prepare-container.sh
@@ -39,3 +39,14 @@ EOF
 fi
 
 apt dist-upgrade
+
+eatmydata_check=$(dirname "$0")/.use-eatmydata
+if [ "$(cat "${eatmydata_check}" || true)" == "true" ]; then
+  # Install and use eatmydata
+  apt install eatmydata
+
+  mkdir -p "/usr/local/libexec/dt-deb-action"
+  ln -s /usr/bin/eatmydata /usr/local/libexec/dt-deb-action/dpkg
+  echo 'Dir::Bin::dpkg /usr/local/libexec/dt-deb-action/dpkg;' > \
+    /etc/apt/apt.conf.d/00dt-deb-action-eatmydata
+fi
--- a/gh-actions/common/build-debian/prepare-container.sh
+++ b/gh-actions/common/build-debian/prepare-container.sh
@@ -1,6 +1,41 @@
 #!/usr/bin/env bash
 set -eux
 
+# Disable installing of manpages and docs
+cat <<"EOF" | tee /etc/dpkg/dpkg.cfg.d/01_nodoc
+# Delete man pages
+path-exclude=/usr/share/man/*
+
+# Delete docs
+path-exclude=/usr/share/doc/*
+path-include=/usr/share/doc/*/copyright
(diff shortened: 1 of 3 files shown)


3v1n0:
@didrocks can you give this a look please? :)

gh-actions/common/build-debian/prepare-container.sh
```diff
@@ -0,0 +1,52 @@
...
+
+  locales_needed=false
+  if echo "${build_deps}" | grep -Fqs language-pack ||
+      echo "${build_deps}" | grep -Fqs locales; then
+      locales_needed=true
+  fi
+
+  # We can't use ${GITHUB_ENV} here, because the github ${{ env.XXX }} variables
+  # arent't visible here, and we don't want to expose each one at call time.
+  echo "$locales_needed" > "${locales_required_check}"
```
didrocks, line 30: nitpick: missing `${locales_needed}`

gh-actions/common/build-debian/prepare-container.sh
```diff
@@ -0,0 +1,52 @@
...
+path-exclude=/usr/share/doc/*
+path-include=/usr/share/doc/*/copyright
+EOF
+
+echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90aptyes
+apt update
+
+locales_required_check=$(dirname "$0")/.locales-required
+if ! [ -e "${locales_required_check}" ] && [ -f ./debian/control ]; then
+  apt install dctrl-tools
```
didrocks, line 19: didn’t we say we always use `-y`?

gh-actions/common/build-debian/prepare-container.sh
```diff
@@ -0,0 +1,52 @@
...
+
+if [ "$(cat "${locales_required_check}" || true)" != "true" ]; then
+  # Disable installing of locale files
+  cat <<"EOF" | tee /etc/dpkg/dpkg.cfg.d/01_nolocales
+# Delete locales
+path-exclude=/usr/share/locale/*
+EOF
+fi
+
+apt dist-upgrade
```
didrocks, line 41: ditto on "yes" by default?

didrocks (changes requested):
I only have 3 small nitpicks, but should be good to go then!

gh-actions/common/build-debian/prepare-container.sh
3v1n0: Well yeah... I had avoided it through env variables though, but let me do it...

3v1n0: Actually it would be pretty dumb, because exactly to avoid this to be forgotten I did initially:

`echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90aptyes`

didrocks: You’re correct

didrocks (approved):
Good work!

3v1n0:
Cool, FYI I've dropped the last commit since it got upstreamed at https://github.com/canonical/setup-lxd/pull/25 :)

PR #59, opened 2025-01-06
fix(gotestfmt): Fix incorrect gotestfmt output

adombeck:
Use our fork until https://github.com/GoTestTools/gotestfmt/pull/62 has been merged by upstream.

UDENG-5451

### DIFF
--- a/gh-actions/go/gotestfmt/action.yaml
+++ b/gh-actions/go/gotestfmt/action.yaml
@@ -13,7 +13,7 @@ runs:
       run: |
         # Install gotestfmt
         set -eu
-        TOOL=github.com/gotesttools/gotestfmt/v2/cmd/gotestfmt
+        TOOL=github.com/ubuntu/gotestfmt/v2/cmd/gotestfmt
         VERSION=$(go mod edit --json | jq -r ".Require[] | select(.Path==\"${TOOL}\") | .Version" || true)
 
         if [ -z "${VERSION}" ]; then


gh-actions/go/gotestfmt/action.yaml
```diff
@@ -13,7 +13,7 @@ runs:
       run: |
         # Install gotestfmt
         set -eu
-        TOOL=github.com/gotesttools/gotestfmt/v2/cmd/gotestfmt
+        TOOL=github.com/adombeck/gotestfmt/v2/cmd/gotestfmt
```
3v1n0, line 16: Maybe use it from a specific branch? So that your fork can change without breaking the actions?

Or what about moving it to a canonical-owned fork instead? As I don't feel `gotestfmt` will handle this very soon...

adombeck: > Or what about moving it to a canonical-owned fork instead? As I don't feel gotestfmt will handle this very soon...

Yeah I also thought about that. Works for me, but I don't think I have the permission to create a fork under the canonical namespace

adombeck:
@didrocks ping

didrocks:
Should we move it to the ubuntu org maybe? If so, I’m happy to handle this.

adombeck:
> Should we move it to the ubuntu org maybe? If so, I’m happy to handle this.

Sure, thanks!

gh-actions/go/gotestfmt/action.yaml
adombeck: that's done now

didrocks (approved):
Thanks a lot!

PR #60, opened 2025-01-14
feat(gh-actions): Add macOS compatibility for go code-sanity and gotestfmt

hk21702:
Updates to add macOS compatibility for go code-sanity and gotestfmt

- Use brew to install jq in code-sanity
- Remove and replace unsupported flags in the install command for gotestfmt
  - Use mkdir instead of -D
  - Remove -t
  - Use -m instead of --mode
  
Minor spelling fix

---
UDENG-5777


didrocks (approved):
Looks good to me, easy and great that this is enough to support Mac OS! :)

Nice work!

PR #61, opened 2025-01-23
maint(gh-actions/run-autopkgtests): Use main branch for setup-lxd

3v1n0:
It contains the fix to speedup the image creation, and we can trust that main will be well managed.

If not we can always go back to a version we know it works.


3v1n0:
@didrocks easy karma :)

didrocks (approved):
indeed, easy! :)

PR #62, opened 2025-01-24
maint(test-build-deb): Test skipping linitian o source build

3v1n0:
We're doing it in other jobs but we didn't test the skip case


didrocks (approved):
Let's go with this!

PR #67, opened 2025-04-07
fix(rust/code-sanity): Fix "rustfmt-check" step

adombeck:
The step succeeded even if the Rust code is not formatted correctly. That's because running `cargo fmt` formats the code and exits with 0. We want `cargo fmt --check` instead, which doesn't actually format the code but checks if it needs formatting, and exits with 1.


adombeck:
@didrocks mind reviewing this one?

didrocks (approved):
Easy one! Thanks for looking at this :)

didrocks (approved):

PR #69, opened 2025-07-17
maint: Allows supplying a GH token

cnihelton:
`arduino/setup-protoc` allows passing a `GITHUB_TOKEN` to authenticate API requests against GitHub in order to prevent API rate limits.
This change set aims to receive and propagate the token until it reaches that action.

Temporarily we need to point to an internal action to a branch, but after merging I'll have to come back to it and retarget main.

When using this token, make sure to limit the permissions to the minimum possible.  [This run](https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/16355485002/job/46212676558?pr=1256) for example is running with empty permissions ( `permissions: {}` ).

### DIFF
--- a/gh-actions/common/gh-perms/action.yaml
+++ b/gh-actions/common/gh-perms/action.yaml
@@ -36,7 +36,7 @@ runs:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         set -eu
-        if [ `gh pr comment -b "You should restrict the github token permissions of your jobs (to none if possible `permissions: {}`)" ${{ github.event.pull_request.number }}`]; then
+        if [ `gh pr comment -b "You should restrict the github token permissions of the ${{ github.job }} job (ideally to none - `permissions: {}`)" ${{ github.event.pull_request.number }}`]; then
           echo "::error::GITHUB_TOKEN has 'pull-requests:write' permission, but it should be 'none' or 'read'."
           echo "status=failed" >> $GITHUB_OUTPUT
         fi


CarlosNihelton:
`permissions` is a child of `jobs` and `job` only. Since actions don't create jobs they don't have a way to set permissions. I'm not aware of any API that would allow that to be made programmatically.

It seems that actions can indeed access the token implicitly via the `${{ github.token }}` context variable (without the caller job passing it as an input). [Documentation says](https://docs.github.com/en/actions/tutorials/use-github_token-in-workflows#using-the-github_token-in-a-workflow):

> An action can access the GITHUB_TOKEN through the `github.token` context even if the workflow does not explicitly pass the GITHUB_TOKEN to the action. As a good security practice, you should always make sure that actions only have the minimum access they require by limiting the permissions granted to the GITHUB_TOKEN. 

I'm not a big fan of this silent pass because if the caller doesn't pay attention to restrict the permissions (what is the norm in GitHub Actions despite recommendations) a downstream action can do whatever with the caller's token without permission. I prefer the verbose input passing than the silent one.

I guess `arduino/setup-protoc` developers agree with me otherwise they'd always do authenticated requests.

In practice we should always control and restrict our token permissions on all jobs for security measures.

hk21702:
I think you're correct; I was mixing reusable workflows and actions in my head. 

There isn't a method of restricting permissions within the actions file itself, so what you have currently is probably the safest method since the only other option would be a silent pass.

gh-actions/go/code-sanity/action.yaml
```diff
@@ -16,7 +16,9 @@ inputs:
   generate-diff-paths-to-ignore:
     description: Files and paths to ignore when checking if generated files changed. This is passed to git update-index.
     default: po/* docs/**/*.md README.md
-
+  token:
+    description: A token to authenticate GH API calls with. It will be passed to setup-protoc for example. Make sure to limit the permissions.
```
hk21702, line 20: Nitpick: It might be beneficial to note that a zero-permissions token is acceptable for what the token is currently being used for. Should the token ever be used for something else in the future and that requires additional permissions, we'd update this description. This is to just give the user an idea of how much they can limit permissions.

gh-actions/go/code-sanity/action.yaml
```diff
@@ -141,12 +143,12 @@ runs:
       if: ${{ always() && steps.go-version.outcome == 'success' }}
       id: go-generate-check
       # https://github.com/orgs/community/discussions/25246
-      uses: canonical/desktop-engineering/gh-actions/go/generate@main
+      uses: canonical/desktop-engineering/gh-actions/go/generate@setup-protoc-token
```
hk21702, line 146: It might not be necessary to have this temporary change here.

For the workflows that you currently have relying on this new setup, I would just point them to a fork temporarily instead of the `setup-protoc-token` branch.

My main worry here is just that if the `setup-protoc-token` branch gets deleted either accidentally or by a repository setting once this PR gets merged, then everything relying on this action would break until it gets reverted.

hk21702 (approved):
Seems good to me!

As far as I can tell, with how default is being used, existing usages shouldn't break, but maybe it would be best to wait for Monday at least before merging still.

didrocks (commented):
To continue on the topic of tokens: can we envision a way to ensure that the given token has restricted permissions, so not downgrading it, but just erroring out if this gives too many permissions? I’m not a big fan of passing such a token to the action (which can still retrieve it, as you say though…).

CarlosNihelton:
There is no clean way, meaning no API endpoint we could call to introspect or at least validate the token permissions directly. But we could go the indirect way, attempting actions that should not be allowable. Using the GitHub CLI is quite straightforward to implement something, but it's tedious to make it comprehensive, as it will be based on trying and failing for each one of the possible permissions below:

```yaml
permissions:
  actions: read|write|none
  attestations: read|write|none
  checks: read|write|none
  contents: read|write|none
  deployments: read|write|none
  id-token: write|none
  issues: read|write|none
  models: read|none
  discussions: read|write|none
  packages: read|write|none
  pages: read|write|none
  pull-requests: read|write|none
  security-events: read|write|none
  statuses: read|write|none
```

We can pick key actions to try and fail if those succeed (like creating PR comments or creating commits). WDYT?

didrocks:
yes! Let’s pick some we want to try and be done with it!

CarlosNihelton:
When working with private repositories, indeed setting  `permissions: {}` makes a huge difference in what information can be retrieved from a repo by an action token.

image

[Full logs of the run above: logs_42248057033.zip](https://github.com/user-attachments/files/21395706/logs_42248057033.zip)

For an open source repo, most stuff can be read, such as issues, PRs, security events and such, so testing for permissions would require exercising the 'write' permissions. That can be disturbing, side effects like new random issues and discussions being created. So instead, I'm placing just the most local side effect checks I could think of and emphasize the ideal case of no permissions. I also preserve the callers right to opt in/out of individual checks.

didrocks (approved):
Thanks for doing the change!

PR #72, opened 2025-07-25
fix: Misc improvements in gh-perms action

cnihelton:
There were a few typos and some misleading shell backticks. More importantly this PR adds a test case for that action as well as preventing surprising side effects, like leaking a commit that would never be pushed to the repo to downstream workflow steps that could compute versions based on Git, for example. We prevent that by doing our own checkout in a clean slate, where we ensure the credentials well be persisted, so git push would just work if the token allows it to work. We strive to checkout just the branch to prevent checking the entire repository's history. Still fetches more stuff than just a single detached head commit but less than the `fetch-depth: 0` option, just enough to make git push work.

### DIFF
--- a/gh-actions/common/gh-perms/action.yaml
+++ b/gh-actions/common/gh-perms/action.yaml
@@ -32,7 +32,7 @@ runs:
       env:
         GH_TOKEN: ${{ inputs.token }}
         REPO_PATH: ${{ github.repository }}
-        NEW_BRANCH: ${{ format('check_push-deleteme-{0}', github.run_id) }}
+        NEW_BRANCH: ${{ format('check_push-deleteme-{0}_{1}', github.run_id, github.sha) }}
       run: |
         set -eu
         git config user.name github-actions[bot]


gh-actions/common/gh-perms/action.yaml
```diff
@@ -9,26 +9,50 @@ inputs:
...
+      working-directory: '.source-checkout-${{ github.run_id }}'
+      env:
+        GH_TOKEN: ${{ inputs.token }}
+        REPO_PATH: ${{ github.repository }}
+        REF_NAME: "refs/heads/${{ github.head_ref }}"
       run: |
         set -eu
         git config user.name github-actions[bot]
         git config user.email github-actions[bot]@users.noreply.github.com
+        # Still fetches less stuff then checking out with fetch-depth: 0.
```
hk21702, line 38: Maybe we can skip the pull if ref is passed through the checkout action?

https://github.com/actions/checkout#push-a-commit-to-a-pr-using-the-built-in-token

hk21702 (commented):
I'm not entirely sure about my comments, but they might be worth discussing more.

gh-actions/common/gh-perms/action.yaml
```diff
@@ -38,13 +62,13 @@ runs:
         GH_TOKEN: ${{ inputs.token }}
       run: |
         set -eu
-        if [ `gh pr comment -b "You should restrict the github token permissions of the ${{ github.job }} job (ideally to none - `permissions: {}`)" ${{ github.event.pull_request.number }}`]; then
-          echo "::error::GITHUB_TOKEN has 'pull-requests:write' permission, but it should be 'none' or 'read'."
+        if gh pr comment -b "You should restrict the github token permissions of the '${{ github.job }}' job (ideally to none - 'permissions: {}')" ${{ github.event.pull_request.number }}; then
```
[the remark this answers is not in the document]
CarlosNihelton: Yeah, this should be skipped if not under a PR, since it's testing the token's permission to create or change PRs and we don't want to create trash in the callers repositories.

gh-actions/common/gh-perms/action.yaml
CarlosNihelton: Very good suggestion indeed. It's much cleaner.

hk21702 (approved):
Looks good to me overall!

From a fresh look, I do wonder if it would be even safer for the `contents: write` check to attempt to create a new branch instead of pushing to the existing ref. Then, if we're certain that we've created a new branch, we can even cleanup and delete it afterward.

CarlosNihelton:
> create a new branch instead of pushing to the existing ref. Then, if we're certain that we've created a new branch, we can even cleanup and delete it afterward.

Yeap, it seems possible and a good idea!

gh-actions/common/gh-perms/action.yaml
```diff
@@ -9,42 +9,73 @@ inputs:
...
+      working-directory: '.source-checkout-${{ github.run_id }}'
+      env:
+        GH_TOKEN: ${{ inputs.token }}
+        REPO_PATH: ${{ github.repository }}
+        NEW_BRANCH: ${{ format('check_push-deleteme-{0}', github.run_id) }}
       run: |
         set -eu
         git config user.name github-actions[bot]
         git config user.email github-actions[bot]@users.noreply.github.com
+        git checkout -b "check-push-${{ github.run_id }}"
```
hk21702, line 40: I don't think this matters in terms of functionality unless the ref branch happens to match `NEW_BRANCH`, but should the branch here be `NEW_BRANCH`?

gh-actions/common/gh-perms/action.yaml
```diff
@@ -9,42 +9,73 @@ inputs:
...
+        if git push --set-upstream origin "${NEW_BRANCH}"; then
+          echo "::error::'token' has 'contents:write' permission, but it should be 'none' or 'read'."
           echo "status=failed" >> $GITHUB_OUTPUT
         fi
+    - name: Clean up temporary branch
+      working-directory: '.source-checkout-${{ github.run_id }}'
+      shell: bash
+      if: ${{ steps.check_push.outputs.status == 'failed' }}
+      env:
+        NEW_BRANCH: ${{ format('check_push-deleteme-{0}', github.run_id) }}
```
hk21702, line 53: Nitpick, but to keep it so this is only defined once, you could probably output it from the prior step and use that here.

hk21702 (commented):
Just a few nitpicks!

Also, for extra safety, we can maybe attempt to fetch `NEW_BRANCH` from remote? I don't think it matters all that much in the end given how `NEW_BRANCH` is set currently, but there is always a nervous feeling when potentially pushing and deleting things.

gh-actions/common/gh-perms/action.yaml
CarlosNihelton: Indeed!

hk21702 (approved):
Seems good to me following the new changes!

CarlosNihelton:
Time to invoke @didrocks !

PR #73, opened 2025-08-28
feat(gotestfmt): Divert "build-output" actions

hk21702:
Currently, `gotestfmt` does not know how to handle outputs from `go test` where the `Action` field is "build-outputs". If it receives a line of this type, then it will panic and print out a byte-encoded string representation of that input's `Output` field.

From what I've seen, this kind of output usually occurs due to a warning output from an underlying toolchain like GCC when using CGo. They can be suppressed with `CGO_LDFLAGS="-w"` but this resolution isn't always intuitive due to how the issue manifests itself. Additionally, these warnings may be genuinely useful.

This PR diverts lines where the `Action` field is "build-outputs", wrapping the line with some additional context and sending it to stderr instead of allowing it to be passed forwards to `gotestfmt` which would in turn panic.

For an example of these changes in use, see this workflow run (ignore the one failure, GitHub's API was degraded at the time): https://github.com/ubuntu/ubuntu-insights/actions/runs/17282035065/job/49052211644?pr=231

### DIFF
--- a/gh-actions/go/gotestfmt/gotestfmt
+++ b/gh-actions/go/gotestfmt/gotestfmt
@@ -62,6 +62,17 @@ strip_empty_coverage_lines() {
     jq -c 'select(.Output != null and (.Output | contains("coverage: 0.0%") | not))'
 }
 
+divert_build_output_action_lines() {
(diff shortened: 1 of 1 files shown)


CarlosNihelton (approved):
🤯

PR #75, opened 2025-08-31
Update golangci-lint-action to v8

adombeck:
Since v7, golangci-lint-action supports golangci-lint v2, and doesn't support v1 anymore. So this is a prerequisite to be able to migrate our projects to golangci-lint v2, but it also requires us to do so.

UDENG-7801


adombeck:
@didrocks for context, see the description of https://warthogs.atlassian.net/browse/UDENG-7800

adombeck:
> Excellent! Thanks for tackling this!

:heart: 

> As part of the request there is use that tag in all repos which currently use the main branch (this GitHub search shows 10 repos), I suggest that we use that opportunity to rather migrate them directly to golangci-lint config v2 in them?

The problem is that those repos use the main branch of this GitHub Action, and  golangci-lint config v2 is incompatible with golangci-lint v1. So to avoid breaking their CI, we can't merge this PR as long as any repos still target main, and we need to make the switch to using golangci-lint v2 and migrating the config in the same PR (in the respective repo).

We could avoid the tags if we first make all repos use the commit ID of current main, then merge and this and create PRs to both migrate the config and let them target main again (for [authd a PR is ready](https://github.com/ubuntu/authd/pull/1049), but it required quite a few changes to fix or silence all the issues found by golangci-lint v2) . The benefit of using a tag instead of a commit is that if any other changes to this GitHub Action are merged before we were able to migrate all to repos to v2, we could move the tag forward. It would also make it easier to handle these situations next time we need to merge a breaking change, because then all repos already use a tag, so we don't need to make all of them use a different commit before we can merge the breaking changes.

didrocks (approved):
> The problem is that those repos use the main branch of this GitHub Action, and golangci-lint config v2 is incompatible with golangci-lint v1. So to avoid breaking their CI, we can't merge this PR as long as any repos still target main, and we need to make the switch to using golangci-lint v2 and migrating the config in the same PR (in the respective repo).

Right right, I got that part :) My suggestion was also to move to golang-ci v2 those projects as you have to get into them one by one already, but I understand that could be too much.

You convince me on the tag vs version. So let’s use version and not be afraid on forever increasing version + retagging vX when needed (as you need to retag everytime to follow a version, GH Action is not really correct with that).

Meanwhile, I’m approving even if we need to hold on on merging though.

PR #78, opened 2025-09-03
Install latest protoc

adombeck:
WIP

### DIFF
--- a/gh-actions/go/generate/action.yaml
+++ b/gh-actions/go/generate/action.yaml
@@ -17,7 +17,7 @@ inputs:
     description: Fail the job on any detected diff.
     default: "true"
   token:
-    description: A token to authenticate GH API calls with. It will be passed to setup-protoc for example. Make sure to limit the permissions, e.g. a zero-permissions token is acceptable.
+    description: A token to authenticate GH API calls with. It will be used to get the latest protoc release, for example. Make sure to limit the permissions, e.g. a zero-permissions token is acceptable.
     default: ""
 outputs:
   diff:
@@ -52,10 +52,22 @@ runs:
       with:
         token: ${{ inputs.token }}
     - name: Install latest protoc
-      uses: arduino/setup-protoc@v3
       if: ${{ steps.proto-deps.outputs.needs-protoc == 'true' }}
       with:
         repo-token: ${{ inputs.token }}
+      shell: bash
+      run: |
+        AUTH_HEADER=""
+        if [ -n "${{ inputs.token }}" ]; then
+          AUTH_HEADER="-H 'Authorization: Bearer ${{ inputs.token }}'"
+        fi
+        PROTOC_VERSION=$(eval curl -s $AUTH_HEADER \
+          https://api.github.com/repos/protocolbuffers/protobuf/releases/latest \
+          | jq -r '.tag_name' | sed 's/^v//')
+        curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-linux-x86_64.zip
+        unzip -o protoc-${PROTOC_VERSION}-linux-x86_64.zip -d $HOME/.local
+        echo "$HOME/.local/bin" >> $GITHUB_PATH
+
     - name: Regenerate files with go generate
       id: go-generate-run
       working-directory: ${{ inputs.working-directory }}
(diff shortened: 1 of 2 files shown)


gh-actions/go/generate/action.yaml
```diff
@@ -52,10 +52,21 @@ runs:
       with:
         token: ${{ inputs.token }}
     - name: Install latest protoc
-      uses: arduino/setup-protoc@v3
       if: ${{ steps.proto-deps.outputs.needs-protoc == 'true' }}
-      with:
-        repo-token: ${{ inputs.token }}
+      run: |
+        AUTH_HEADER=""
```
didrocks, line 57: mine using a `set -eu` on top?

gh-actions/go/generate/action.yaml
```diff
@@ -52,10 +52,21 @@ runs:
...
+        if [ -n "${{ inputs.token }}" ]; then
+          AUTH_HEADER="-H 'Authorization: Bearer ${{ inputs.token }}'"
+        fi
+        PROTOC_VERSION=$(eval curl -s $AUTH_HEADER \
```
didrocks, line 61: I don’t think you did that without research, but why do we need the AUTH_HEADER? What happens if it’s empty as there is one code path without it?

General thing is that I would prefer we avoid requiring it (for forks in particular) if possible. This was my main grip with the arduino action: we needed to pass it.

gh-actions/go/generate/action.yaml
```diff
@@ -52,10 +52,21 @@ runs:
...
+          https://api.github.com/repos/protocolbuffers/protobuf/releases/latest \
+          | jq -r '.tag_name' | sed 's/^v//')
+        curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-linux-x86_64.zip
+        unzip -o protoc-${PROTOC_VERSION}-linux-x86_64.zip -d $HOME/.local
+        rm protoc-${PROTOC_VERSION}-linux-x86_64.zip
+        echo "$HOME/.local/bin" >> $GITHUB_PATH
+      shell: bash
```
didrocks, line 68: nitpick: I think in most actions, we pass the shell before `run`, just look at other actions and please align. If we are not consistent, my bad for letting that pass through :)

didrocks (changes requested):
I will be really happy once we remove this action because I never really like it when it was introduced! Thanks for tackling this.

I have some questions and one change request!

gh-actions/go/generate/action.yaml
adombeck: Using the GitHub token in the request to the GitHub API endpoint avoids rate limiting. The `arduino/protoc` action used it for the same purpose, and it did not actually require it, but we did recently pass it with 895c1139e7452498a8155c5e352843a6eb307faa. I don't want to break things for anyone, so I think its better to keep using it.

gh-actions/go/generate/action.yaml
adombeck: done

gh-actions/go/generate/action.yaml
didrocks: Indeed, I can see that we can hit the limit quite easily, thanks for looking at this! It makes sense to keep this logic then.

Maybe just a warning/notice to be printed in the GH summary in case the download fails and we have an empty `AUTH_HEADER` with instructions to pass the token? That would help debugging.

gh-actions/go/generate/action.yaml
didrocks: That would be excellent! Hopefully, I was consistent in my own actions :)

I agree with a commit and PR to align all of this, include `set -eu` and `name of steps`!

gh-actions/go/generate/action.yaml
didrocks: (you didn’t push it yet ;-))

gh-actions/go/generate/action.yaml
adombeck: Okay, see you on https://github.com/canonical/desktop-engineering/pull/81 :)

gh-actions/go/generate/action.yaml
didrocks: Saw it :)

gh-actions/go/generate/action.yaml
adombeck: done, please take a look

adombeck:
rebased on main

PR #79, opened 2025-09-04
feat: Add GitHub Action common/dpkg-exclude-locales-and-manpages

adombeck:
This GitHub Action configures dpkg to exclude locales and manpages from installed packages, speeding up package installation. We should use it in all jobs which use APT to install packages, because it can significantly speed up the installation time. Avoiding the man-db triggers alone can reduce the installation time by up to two minutes.

UDENG-7865

### DIFF
--- a/gh-actions/common/dpkg-exclude-locales-and-manpages/action.yml
+++ b/gh-actions/common/dpkg-exclude-locales-and-manpages/action.yml
@@ -1,0 +1,16 @@
+name: Exclude locales and manpages from dpkg
+description: Configures dpkg to exclude locales and manpages from installed packages, speeding up package installation.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Exclude locales and manpages from dpkg
+      id: exclude-locales-and-manpages
+      shell: bash
+      run: |
+        cat <<EOF | sudo tee /etc/dpkg/dpkg.cfg.d/01-exclude-locales-and-manpages
+        path-exclude=/usr/share/locale/*
+        path-exclude=/usr/share/man/*
+        path-exclude=/usr/share/doc/*
+        path-exclude=/usr/share/info/*
+        EOF
--- a/gh-actions/common/dpkg-install-speedup/action.yml
+++ b/gh-actions/common/dpkg-install-speedup/action.yml
@@ -1,4 +1,4 @@
-name: Exclude locales and manpages from dpkg
+name: Speed up dpkg package installation
 description: Configures dpkg to exclude locales and manpages from installed packages, speeding up package installation.
 
 runs:


adombeck:
Tested in https://github.com/ubuntu/authd/actions/runs/17468260999/job/49609652710?pr=1062#step:2:1.

Note how "Processing triggers for man-db" took less than a second in that run:

image

While it can take up to two minutes in other runs, for example https://github.com/ubuntu/authd/actions/runs/17432755496/job/49494832954#step:2:172:

image

didrocks:
If we want to really speed up apt through dpkg package installation, I suggest we rather use something to disable `fsync` or `eatmydata` rather than excluding specific files with a tons of actions, wdyt?

adombeck:
> So, there is the potential for a lot of specific actions to be added and then included in each project's workflows (and we don’t wait to include them all by default, before, for instance, we may in one project rely on having the manpage). This is this matrix that I am a little bit scared of compared to disabling fsync or something like that that can be applied transparently and easily. Do you see what I mean?

I think I understand your concern about the effort to add this (and potentially other) actions to many GitHub workflows. What I don't really see is how any other approach, like eatmydata, would improve things. We would also have to manually add that to all workflows where we want to use it, right? Is it that this could be done in the GitHub workflows directly instead of a separate action?

> Just tell me about the results you are seeing with eatmydata or just fsync disablement on the system, and I think then we can reassess!

I'll do these assessments now and report back.

adombeck:
> I'll do these assessments now and report back.

Unfortunately, disabling fsync via eatmydata does not seem to significantly reduce the time of the long "Processing triggers for man-db" operation:

image

adombeck:
>  I think maybe thus having a general "install speedup" action where you can enable/disable manpage strips (and thus, we build from it with more features on by default) would be a solution, what do you think?

Yes, one configurable "APT/dpkg install speedup" action sounds better than five separate ones. I will rename it. For now, there's not really anything to configure, unless we already want the exclusion of locales (`/usr/share/locale/*`) to be separate from man pages and docs (`/usr/share/man/*`, `/usr/share/doc/*`, `/usr/share/info/*`). Should I do that?

adombeck:
> we will need to add maybe 5 of them to make it a decent speed about it

Let me add: I agree that for packages which trigger an initramfs update, avoiding that will also bring a nice speed-up. However, as shown in the PR description, preventing the man-db triggers alone can save up to 2 minutes, which is already a decent speed-up IMO, and makes the action worth it without any additions.

adombeck:
I renamed the action to `dpkg-install-speedup`. @didrocks, do you think that's generic enough or should I also omit the `dpkg` part?

didrocks (approved):
It could be apt too, but I guess I’m fine with that as it’s all under the "package" mentra.

Thanks for changing it!

3v1n0:
Something similar was used already in https://github.com/canonical/desktop-engineering/blob/main/gh-actions/common/build-debian/prepare-container.sh not sure if we can factorize some things

PR #86, opened 2025-09-15
feat: gh-actions/build-debian: Add extra-apt-repositories input

adombeck:
When building authd, we need to add a PPA before the build dependencies are installed.

UDENG-8333

### DIFF
--- a/gh-actions/common/build-debian/prepare-container.sh
+++ b/gh-actions/common/build-debian/prepare-container.sh
@@ -14,6 +14,11 @@ EOF
 echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90aptyes
 apt update
 
+# Install ca-certificates to ensure that APT can access HTTPS repos
+apt install ca-certificates
+# Update package lists again, to ensure that we also have lists from HTTPS repos
+apt update
+
 locales_required_check=$(dirname "$0")/.locales-required
 if ! [ -e "${locales_required_check}" ] && [ -f ./debian/control ]; then
   apt install dctrl-tools
--- a/gh-actions/common/build-debian/action.yaml
+++ b/gh-actions/common/build-debian/action.yaml
@@ -21,6 +21,11 @@ inputs:
     # FIXME: this should default to '', but we don't want to break job depending on us for now
     default: 'ca-certificates'
 
+  extra-apt-repositories:
+    description: |
+      A list of extra APT repositories to add.
+      The entries are passed to add-apt-repository, one per line.
+
   extra-source-build-script:
     description: |
       A script to run to prepare the source build machine.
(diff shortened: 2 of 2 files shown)


adombeck:
Tested in https://github.com/ubuntu/authd/actions/runs/17766612028/job/50491504238?pr=1074#step:3:2913

3v1n0:
You may need to use https://github.com/canonical/desktop-engineering/blob/c5e282e9a1a3b293a824481a8c6081487b046100/.github/workflows/test-build-deb.yaml#L165

However, in general while installing few things before the `source` package is built is kinda acceptable, I don't think we should allow to do **anything** during `Build packages` phase, or we'd end up polluting the build with stuff that is not part of debian/rules.

Maybe we should just switch here to use `sbuild` instead (in a container - I had planned it for some long time, TBH), now that it can create build roots automatically.

The only thing we could IMHO, is the support for adding `extra-apt-sources` in the same way we do it for `autopkgtests`, so ideally we should just modify `/etc/apt/sources.list.d/`, either manually in the script or we could do this by: run another image instance, add the repository using `add-apt-repository`, copy the content of `/etc/source.list.d` (and the keys if any) to the build image.

But I think we should not do anything that is not in `debian/rules` in the build phase, that was the main point of this action after all.

didrocks:
I think I understand @3v1n0's concern and I agree that in general, we should not encode additional commands manually in our CI workflow as it might be manual steps to be done on everyone's machine to prepare the source package (while in general, we should have this in debian/rules as we did with autovendoring the dependencies).

This is why the idea to offer adding a ppa is fine, because restrictive, rather than letting shell scripts to be executed. I will let @3v1n0 confirms that this is what he means…

adombeck:
Tested in https://github.com/ubuntu/authd/actions/runs/18009070882/job/51237405066?pr=1074#step:3:4218

didrocks (approved):
It fully does for me, thanks for doing that change!

adombeck:
> As per https://github.com/ubuntu/authd/actions/runs/18524287925/job/52791286242#step:3:3995 we may also need ca-certificates being installed

yeah that might be the issue

adombeck:
It's working now

gh-actions/common/build-debian/prepare-container.sh
```diff
@@ -14,6 +14,11 @@ EOF
 echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90aptyes
 apt update
 
+# Install ca-certificates to ensure that APT can access HTTPS repos
+apt install ca-certificates
+# Update package lists again, to ensure that we also have lists from HTTPS repos
+apt update
```
3v1n0: Oh @adombeck , I would have preferred this to happen *only* if we have PPAs, that's not something we should do by default... The launchpad case is an exception, but otherwise we don't want any network access (it's in fact blocked by iptables) and these packages were installed only in specific cases (when used as input).

adombeck: > The launchpad case is an exception, but otherwise we don't want any network access (it's in fact blocked by iptables) and these packages were installed only in specific cases (when used as input).

I don't understand. We don't have to care about launchpad here because this script is only executed in out GitHub CI, not when building on launchpad, right? In our GitHub CI we do have network access, else all the code around this would also fail (e.g. the `apt update` right above or the `apt dist-upgrade` and `apt install eatmydata` below), right? Which packages were installed in specific cases?

PR #87, opened 2025-09-16
fix: Install govulncheck with the correct Go toolchain version

adombeck:
Unless `GOTOOLCHAIN` is specified, the Go version used by `go install` to build the binary is determined by:
1. The `toolchain` directive in the `go.mod` of the module being installed
2. If there is no `toolchain` directive in the `go.mod` of the module being installed, the system's Go is used (the one in PATH).

We specify the Go version that should be used for tools like govulncheck in `tools/go.mod`. So we need to point `GOTOOLCHAIN` to that to actually use that Go version when building govulncheck.

This avoids errors like:

    Error: /home/runner/go/pkg/mod/golang.org/x/sys@v0.35.0/unix/vgetrandom_linux.go:7:9: file requires newer Go version go1.24 (application built with go1.23)

which occur when govulncheck is built with an older Go version than the required version of the module that it should check.

UDENG-8058

### DIFF
--- a/gh-actions/go/code-sanity/action.yaml
+++ b/gh-actions/go/code-sanity/action.yaml
@@ -162,12 +162,12 @@ runs:
     - name: Install govulncheck
       if: ${{ always() && steps.tooling-version.outcome == 'success' }}
       id: install-govulncheck
-      working-directory: ${{ runner.temp }}
       shell: bash --noprofile --norc -euo pipefail {0}
       run: |
         # Install govulncheck
         echo "::group::Install govulncheck"
-        go install golang.org/x/vuln/cmd/govulncheck@${{ steps.tooling-version.outputs.vulncheck }}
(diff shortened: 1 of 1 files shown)


didrocks (approved):
TBH, I don’t have the whole matrix of cases in mind, but what you did makes sense to me. Maybe we should explore using another approach in the long term like using the base `go.mod` for this.

For now, let’s merge this, thanks!

hk21702:
I think this change is breaking what is present in Ubuntu Insights, where the `go.mod` file in the `tools/` directory does not have a toolchain directive, and the toolchain directive instead lives in a `go.work` file in the root repository.

https://github.com/ubuntu/ubuntu-insights/actions/runs/17783648722/job/50547197132?pr=247

The change will also probably break the Ubuntu Pro for WSL repository since the `go.mod` file in `tools/` also doesn't have a toolchain directive, but that repository is still on v1.

didrocks:
You are right, we should have fallback to go version directive without it. Do you mind handling this @adombeck?

PR #88, opened 2025-09-17
fix: Fix go/code-sanity failing if no toolchain directive in tools/go.mod

adombeck:
The `toolchain` directive in a `go.mod` is optional, but commit 8a48366f25234b5c223db6aac399ec660c4ea8c7 caused the installation of govulncheck to fail with `go: invalid GOTOOLCHAIN "null"` if the directive is missing.

Let's fall back to the Go version specified by the `go` directive in case the `toolchain` directive is missing.


didrocks (approved):
Looks good to me! Let’s wait for Kat doing a testun first before merging it.

PR #91, opened 2025-11-06
feat: Support using go tool directive for protoc deps

adombeck:
* Use `go tool` to check if protoc should be installed.
* Drop support for installing tools from `tools/*.go` files - the only reason we kept it after #89 was to avoid breaking protoc.

### DIFF
--- a/gh-actions/go/generate/action.yaml
+++ b/gh-actions/go/generate/action.yaml
@@ -27,29 +27,15 @@ outputs:
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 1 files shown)


didrocks (commented):
This is indeed way better! Wouldn’t that potentially break some repos though that are depending on this? Should we just update them first?

jibel:
@adombeck can you resolve the conflicts and resubmit for review?

adombeck:
Resolved the conflicts

PR #118, opened 2026-03-20
feat(govulncheck): add input to skip go vulnerability scanning step

gabelluardo:
Add the ability to skip `govulncheck` step.


didrocks (approved):
Thanks a lot! That will be handy until snapd/secboot fix their versionning.

PR #122, opened 2026-05-08
fix(gh-actions/build-debian): Fix deb lines in extra-apt-repositories

adombeck:
'deb ...' lines would be written to `/etc/apt/sources.list` by `add-apt-repository`, which is not shared with later steps (because only `/etc/apt/sources.list.d` is shared).

Write them directly to `sources.list.d` instead.

Also add an input `extra-apt-repositories-pin` to allow prioritizing packages from the extra repos.

### DIFF
--- a/gh-actions/common/build-debian/action.yaml
+++ b/gh-actions/common/build-debian/action.yaml
@@ -227,9 +227,16 @@ runs:
           apt-get update
           apt-get install -y software-properties-common
           echo "${{ inputs.extra-apt-repositories }}" | while read -r repo; do
-          if [ -n "$repo" ]; then
-           add-apt-repository -y "$repo"
-          fi
+            if [ -n "$repo" ]; then
+              if [[ "$repo" == deb* ]]; then
+                # 'deb ...' lines would be written to /etc/apt/sources.list by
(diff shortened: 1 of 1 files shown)


adombeck:
Tested in https://github.com/canonical/authd/actions/runs/25545755437/job/74993600014?pr=1473

gh-actions/common/build-debian/action.yaml
```diff
@@ -219,18 +222,30 @@ runs:
...
           echo "::group::Install extra apt repositories"
           set -euxo pipefail
           apt-get update
           apt-get install -y software-properties-common
           echo "${{ inputs.extra-apt-repositories }}" | while read -r repo; do
-          if [ -n "$repo" ]; then
-           add-apt-repository -y "$repo"
-          fi
+            if [ -n "$repo" ]; then
+              if [[ "$repo" == deb* ]]; then
```
3v1n0, line 236: Should we also support the new format? Not sure how long the old deb lines will be supported

adombeck: not necessary IMO. not even add-apt-repository supports the new format yet AFAICT

3v1n0 (approved):

PR #126, opened 2026-05-21
feat: Allow setting the Team field of new Jira issues

adombeck:
We now require the Team field of issues to be set for them to show up on our pulse board. The jira CLI tool now supports that.

UDENG-10227


didrocks (approved):
Looks good to me! That will be useful

PR #129, opened 2026-06-11
fix: gh-actions/build-debian: Fix 'cd: null directory'

adombeck:
On Resolute, using `cd ''` in bash fails. On Noble it did not.

UDENG-10958

### DIFF
--- a/gh-actions/common/build-debian/action.yaml
+++ b/gh-actions/common/build-debian/action.yaml
@@ -114,7 +114,9 @@ runs:
       run: |
         echo "::group::Create local version with commit and docker container"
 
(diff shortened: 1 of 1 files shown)


adombeck:
Tested in https://github.com/canonical/authd/actions/runs/28530974878?pr=1577

adombeck:
and here is an example of a failure, using the action from current main: https://github.com/canonical/authd/actions/runs/27375873286/job/80899846972?pr=1577

didrocks (approved):
Easy enough to think it’s always a good thing to check before cding into it :)

PR #132, opened 2026-07-23
feat: add action to notify MM about GH failures

hk21702:
Here, we add an action to notify Mattermost about GH failures. When used as intended, it fetches the corresponding run, linking the run, ref, and describing the trigger. It also pings a list of configured users. 

This is pretty similar to what we already use elsewhere. But, instead of a webhook, this utilizes the MM REST API and requires a bot token for use. We also add a few additional features, such as using the repository org/owner icon as the author icon by default.

In terms of safety, as long as the MM token is implemented using a repository secret, it should be that only actors with write permissions can send notifications. That being said, caution should still be taken when using this action in public repositories.

Example notification:
image


hk21702:
Example run: https://github.com/canonical/desktop-engineering/actions/runs/30026400245/job/89271848221?pr=132

CarlosNihelton (approved):
Beautiful!

PR #133, opened 2026-07-23
feat(g/infra): Add JAAS auth action

hk21702:
Here, we add a JAAS auth action, taking in a client ID and client secret to authenticate with JAAS. By taking advantage of JS action features, we also implement a cleanup hook that clears out the Juju data and logs the user out of JAAS. Though not strictly necessary, it's safer to do this when considering the possibility of non-ephemeral runners.

Note that because of a bug with Juju, the login action requires an additional controller normalization step, where we clear the embedded controller CA certificate from the Juju controllers file. This workaround needs to be enabled until https://github.com/juju/juju/pull/22931 makes it to a stable release.

---


.github/workflows/js-qa.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - "**/*.mjs"
+      - .github/workflows/js-qa.yaml
+
+jobs:
+  check-js-syntax:
+    name: node --check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
```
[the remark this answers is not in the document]
hk21702: The others are out of date.

hk21702:
Example use: https://github.com/canonical/ubuntu-engineering-terraform-models/actions/runs/30057513119?pr=711

.github/workflows/js-qa.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Syntax check all JS files
+        run: |
+          while IFS= read -r file; do
+            echo "Checking $file"
+            node --check "$file"
```
CarlosNihelton, line 29: Shouldn't we have a `package.json` file and use some linter tool beyond just syntax validation?

gh-actions/infra/jaas-auth/dist/main/index.mjs
```diff
@@ -0,0 +1,96 @@
...
+  const controllerHost = required("jaas-controller-host");
+  const clientId = required("juju-client-id");
+  const clientSecret = required("juju-client-secret");
+
+  // Install Juju if needed.
+  try {
+    execFileSync("snap", ["list", "juju"], { stdio: "pipe" });
+  } catch {
+    run("sudo", ["snap", "install", "juju"]);
+  }
```
CarlosNihelton: If the objective is detect missing juju snap and install it we can just `snap install juju` and rely on the fact that it won't fail but rather refresh juju snap if needed or do nothing, right?

```
$ snap install hello  
snap "hello" is already installed, see 'snap help refresh'
$ echo $?                             
0
```

gh-actions/infra/jaas-auth/dist/main/index.mjs
```diff
@@ -0,0 +1,96 @@
...
+  );
+
+  if (input("clear-ca-cert", "true") === "true") {
+    clearCaCert(path.join(jujuData, "controllers.yaml"));
+  }
+
+  fs.appendFileSync(process.env.GITHUB_OUTPUT, `juju-data=${jujuData}\n`);
+  // Persist for the post (cleanup) entrypoint.
+  appendEnv("JAAS_AUTH_JUJU_DATA", jujuData);
+  appendEnv("JAAS_AUTH_CLEANUP", input("cleanup", "true"));
```
CarlosNihelton, line 79: This works because `GITHUB_ENV` is a file, not true environment variables, right?

gh-actions/infra/jaas-auth/dist/post/index.mjs
```diff
@@ -0,0 +1,47 @@
...
+    console.error(
+      `::warning::juju logout failed; removing client state anyway: ${error.message}`,
+    );
+  }
+
+  fs.rmSync(jujuData, { recursive: true, force: true });
+
+  appendEnv("JUJU_CLIENT_ID", "");
+  appendEnv("JUJU_CLIENT_SECRET", "");
+  appendEnv("JUJU_DATA", "");
```
CarlosNihelton, line 44: We're adding more lines to a file here. Shouldn't we remove the lines matching those variable names instead?

CarlosNihelton (commented):
Good job. I only have minor questions, but I'm quite out of my comfort zone, so perhaps not well informed enough to approve this PR right away.

gh-actions/infra/jaas-auth/dist/main/index.mjs
hk21702: Probably yes.

.github/workflows/js-qa.yaml
hk21702: We could maybe use a linter tool, but the JS here is pretty simple. As for a `package.json`, it's not strictly necessary because we don't have any dependencies, and I intentionally avoided them so that we don't have to think about compilation and vendoring, which would then require a much larger headache of how we want to handle that in a monorepo that wasn't designed for it.

gh-actions/infra/jaas-auth/dist/main/index.mjs
hk21702: Yes, `GITHUB_ENV` is a file.

gh-actions/infra/jaas-auth/dist/post/index.mjs
hk21702: I don't think we can remove lines. From what I remember, `GITHUB_ENV` is an append-only log file.

hk21702: https://github.com/actions/runner/issues/1126

gh-actions/infra/jaas-auth/dist/main/index.mjs
CarlosNihelton: Alright, that's something worth doing then.

CarlosNihelton (approved):
I'll let you decide on the snap install thing. Otherwise, LGTM!

didrocks (approved):
Quite similar to Carlos, I’m a little bit too far to get a full view on all the changes, but I didn’t spot anything that was not already mentioned and taken into account.

Good work!

matthew-hagemann (approved):

matthew-hagemann (changes requested):
I've been browsing some of the other teams setups:

## Q1: 
> Though not strictly necessary, it's safer to do this when considering the possibility of non-ephemeral runners

Is this a concern you see in the future? From what I can see in the other repo's, everyone else using runners all seem to be on ephemeral runners across the org, including when running on self-hosted runners. It just adds allot of JS complexity we don't seem to need, no?

## Q2:

> Note that because of a bug with Juju, the login action requires an additional controller normalization step, where we clear the embedded controller CA certificate from the Juju controllers file.

Do we need this? From what I can tell, on PS7, jaas is a trusted cert, so you don't need  `--trust`.  `jaas.ps7.canonical.com`  is served with a public Let's Encrypt cert, so the runner's system CA store validates it automatically. You end up with [so little code,](https://github.com/canonical/platform-engineering-deployments/blob/main/.github/workflows/backup-github-runner-postgresql.yaml#L38-L39) I think we can forgo a dedicated action altogether?


Is there a reason we can't just condense this down to something similar to what PE is doing [here](https://github.com/canonical/platform-engineering-deployments/blob/main/.github/workflows/backup-github-runner-postgresql.yaml#L38-L39)? I don't think we want to make a generalized backup solution, that is more of an IS concern, we just want some interim measure while we wait for support for managed MySQL db's on our infra?

hk21702:
> Is this a concern you see in the future? From what I can see in the other repo's, everyone else using runners all seem to be on ephemeral runners across the org, including when running on self-hosted runners.

Likely no, but it's still a safety mechanism considering that it is direct infrastructure access into our servers. 

Plus, we do in fact have non-ephemeral runners, and I wouldn't be too surprised if one day someone uses them more directly, though its true it's not a concern for the foreseeable future. 

> Do we need this? From what I can tell, on PS7, jaas is a trusted cert, so you don't need  --trust.  jaas.ps7.canonical.com  is served with a public Let's Encrypt cert, so the runner's system CA store validates it automatically.

During my testing, both locally and with the runner, yes, at least for the very first login. Once the fingerprint is trusted, we don't run into it again. It may have changed since then though.

hk21702:
@matthew-hagemann Seems like the requirement for `--trust` and `--no-prompt` might have been related to https://github.com/juju/juju/pull/22931 which recently was fixed. So, they should be removable.

I'd still consider doing the post hook though where there's an explicit log out for safety, though. You never know when there could be a CVE and it's really not that much more complex since we're using plain JS.

With https://github.com/juju/juju/pull/22931 being fixed, we could disable the fix by default, or we can just simplify the action further and remove it entirely. IMO, it's probably better to just remove the dormant code.

hk21702:
New run with changes: https://github.com/canonical/ubuntu-engineering-terraform-models/actions/runs/31398978700/job/93488885157

matthew-hagemann (approved):
One change to remove the unit-tests flow now that their are no tests left.

I'm not personally worried about an issue with cleanup, but its not a huge amount of code, and doesn't hurt.

PR #134, opened 2026-07-23
feat(g/infra): Add backup-database action

hk21702:
Here, we introduce a new action to handle database backups on our infra.

The general flow is selecting a healthy unit of a Juju application and then running a database charm backup action on it. We assume that Juju is already authenticated.

Note that special consideration needs to be made when deciding which unit to run the backup action against. In multi-unit deployments, we typically want to target non-primary units. The exception is with non-TLS deployments, where we must target the primary. With this in mind, we need to have different modes depending on what we would like to do.

However, the only standardized way between the common DB charms, PostgreSQL and MySQL, of determining which unit is the primary is by using `juju status`. Actions like `get-cluster-status` and `get-primary` are not always available.

Additionally, in cases where there is a broken unit, we want to avoid it and not include it for consideration.

For the `dry-run` case, we only run `list-backups`, making no writes.

Note that this action does *not* handle backup expiration. For PostgreSQL, this should be handled by configuring `experimental-delete-older-than-days` with the `s3-integrator` charm. For MySQL, this needs to be handled carefully with S3 lifecycle rules in such a way that `group_replication_id.txt` does not expire.

Refs:
https://canonical.com/data/mysql/docs/8.0/how-to/back-up-and-restore/create-a-backup/
https://canonical.com/data/postgresql/docs/14/how-to/back-up-and-restore/create-a-backup/

---

### DIFF
--- a/.github/workflows/test-backup-database.yaml
+++ b/.github/workflows/test-backup-database.yaml
@@ -1,0 +1,38 @@
+name: Test backup-database
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - gh-actions/infra/backup-database/**
+      - .github/workflows/test-backup-database.yaml
+  pull_request:
+    paths:
+      - gh-actions/infra/backup-database/**
+      - .github/workflows/test-backup-database.yaml
+
+jobs:
+  check-python-syntax:
+    name: py_compile
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Syntax check all Python files
+        run: |
+          while IFS= read -r file; do
+            echo "Checking $file"
+            python3 -m py_compile "$file"
+          done < <(find gh-actions/infra/backup-database -type f -name '*.py' -print0 | sort -z | tr '\0' '\n')
+
+  unit-tests:
+    name: Python unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Run unit tests
+        run: python3 -m unittest discover -s gh-actions/infra/backup-database/tests -p 'test_*.py' -v
--- a/gh-actions/infra/backup-database/README.md
+++ b/gh-actions/infra/backup-database/README.md
@@ -1,0 +1,59 @@
+# Database backup action
+
+This composite GitHub Action selects a healthy unit of a Juju application and
+runs a database charm backup action on it. It requires an already authenticated
+Juju session and reports the result in the GitHub job summary.
+
+The [jaas-auth](../jaas-auth/action.yml) action can be used to install Juju and
+authenticate to JAAS beforehand.
+
+## Inputs
+
+| Input         | Required | Default         | Description                                        |
+| ------------- | -------- | --------------- | -------------------------------------------------- |
+| `model`       | yes      |                 | Juju model name                                    |
+| `model-owner` | no       |                 | Juju model owner, for models owned by another user |
+| `application` | yes      |                 | Application whose units will be considered         |
+| `action`      | no       | `create-backup` | Juju action name                                   |
+| `parameters`  | no       | `{}`            | JSON object containing action parameters           |
+| `unit-role`   | no       | `non-primary`   | `non-primary`, `primary`, or `any`                 |
+| `timeout`     | no       | `6h`            | Duration passed to Juju's `--wait`, e.g. `30m`     |
+| `dry-run`     | no       | `false`         | Skip backup creation but still list backups        |
+
+Only `model` and `application` are strictly required. `model-owner` is only
+needed when targeting a model owned by another user.
+
+## Unit selection
+
+Units with an unhealthy workload (`blocked` or `error`) or agent (`error` or
+`lost`) status are excluded from selection. The unit with the workload status
+message `Primary` is treated as the primary. With `non-primary`, the first
+eligible non-primary unit is selected, falling back to the primary with a
+degraded warning when no eligible replica remains. `any` skips role discovery.
+Excluded units and degraded fallbacks are noted in the job summary.
+
+## Usage
+
+```yaml
+steps:
+  - uses: actions/checkout@v7
+  - name: Authenticate to JAAS
+    id: jaas-auth
+    uses: canonical/desktop-engineering/gh-actions/infra/jaas-auth@main
+    with:
+      jaas-controller: ${{ vars.JUJU_CONTROLLER }}
+      jaas-controller-host: ${{ vars.JUJU_CONTROLLER_HOST }}
+      juju-client-id: ${{ secrets.JUJU_CLIENT_ID }}
+      juju-client-secret: ${{ secrets.JUJU_CLIENT_SECRET }}
+  - uses: canonical/desktop-engineering/gh-actions/infra/backup-database@main
+    env:
+      JUJU_DATA: ${{ steps.jaas-auth.outputs.juju-data }}
+    with:
+      model: example-model
+      model-owner: ${{ vars.JUJU_MODEL_OWNER }}
+      application: database
+      dry-run: "true"
+```
+
+Invoke the action once per backup target. Run with `dry-run: "true"` first to
+validate model access and unit selection before enabling backup creation.
--- a/gh-actions/infra/backup-database/backup.py
+++ b/gh-actions/infra/backup-database/backup.py
@@ -360,7 +360,7 @@ def _select_target_unit(context: RunContext) -> Selection | None:
 
 
 def _run_backup_action(context: RunContext, selection: Selection, notes: str) -> bool:
-    """Run the configured backup action, withholding its captured output."""
+    """Run the configured backup action, withholding its captured stdout."""
 
     target = context.target
     logger.info("Running the configured backup action on %s", selection.unit)
@@ -373,6 +373,7 @@ def _run_backup_action(context: RunContext, selection: Selection, notes: str) ->
         target.action,
         f"--wait={target.timeout}",
         "--format=json",
+        "--quiet",
         *(
             f"{name}={parameter_value(value)}"
             for name, value in target.parameters.items()
@@ -382,6 +383,9 @@ def _run_backup_action(context: RunContext, selection: Selection, notes: str) ->
     if result == 0 and juju_run_succeeded(capture.stdout):
         logger.info("Backup action completed successfully on %s", selection.unit)
         return True
+    stderr = capture.stderr.read_text().strip()
+    if stderr:
+        logger.warning("Backup action stderr: %s", stderr)
     logger.error(
         "Backup action failed for %s on %s; action output was withheld",
         context.target_name,
@@ -409,14 +413,21 @@ def _list_backups(context: RunContext, selection: Selection, notes: str) -> bool
         "list-backups",
         f"--wait={target.timeout}",
         "--format=json",
+        "--quiet",
     ]
     result = capture.run(context.runner, command)
-    capture.print()
-    if result == 0 and juju_run_succeeded(capture.stdout):
+    succeeded = result == 0 and juju_run_succeeded(capture.stdout)
+    if succeeded:
+        capture.print()
         logger.info("Backup listing completed successfully on %s", selection.unit)
         return True
+    stderr = capture.stderr.read_text().strip()
+    if stderr:
+        logger.warning("list-backups stderr: %s", stderr)
     logger.error(
-        "Unable to list backups for %s on %s", context.target_name, selection.unit
+        "Unable to list backups for %s on %s; command output was withheld",
+        context.target_name,
+        selection.unit,
     )
     failure = (
         "list-backups failed during dry run"
--- a/gh-actions/infra/backup-database/backup.py
+++ b/gh-actions/infra/backup-database/backup.py
@@ -1,0 +1,478 @@
+#!/usr/bin/env python3
+
+"""Select a healthy Juju unit and run a database backup."""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+import re
+import tempfile
+from collections.abc import Mapping
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any
+
+from backup_helpers import (
+    CommandCapture,
+    CommandRunner,
+    join_notes,
+    juju_run_succeeded,
+    parameter_value,
+    read_json_object,
+    run_command,
+    set_result_output,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class SelectionError(ValueError):
+    """Raised when no unit satisfies the requested selection policy."""
+
+
+@dataclass(frozen=True)
+class BackupTarget:
+    """Configuration for one database backup target."""
+
+    application: str
+    model: str
+    model_owner: str = ""
+    action: str = "create-backup"
+    parameters: Mapping[str, Any] = field(default_factory=dict)
+    unit_role: str = "non-primary"
+    timeout: str = "6h"
+
+    @property
+    def qualified_model(self) -> str:
+        """Return the owner-qualified model name expected by Juju."""
+
+        if not self.model_owner:
+            return self.model
+        return f"{self.model_owner}/{self.model}"
+
+
+@dataclass(frozen=True)
+class Exclusion:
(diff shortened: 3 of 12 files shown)


gh-actions/infra/database-backup/backup.py
```diff
@@ -0,0 +1,459 @@
...
+        model=values.get("MODEL", ""),
+        application=values.get("APPLICATION", ""),
+        model_owner=values["MODEL_OWNER"],
+        action=values.get("ACTION", "create-backup"),
+        parameters=parameters,
+        unit_role=values.get("UNIT_ROLE", "non-primary"),
+        timeout_minutes=values.get("TIMEOUT_MINUTES", "360"),
+    )
+
+
```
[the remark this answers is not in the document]
hk21702: Technically, if Juju is already switched to the correct model/controller, then we don't need to pass in the model, or model owner. Application is the only strictly required input. However, when we do have a model owner, then the model becomes strictly required as well. Model owner may not be strictly required in some scenarios when model is provided, mostly in local scenarios.

gh-actions/infra/backup-database/README.md
```diff
@@ -0,0 +1,56 @@
+# Database backup action
+
+This composite GitHub Action selects a healthy unit of a Juju application and
+runs a database charm backup action on it. It requires an already authenticated
+Juju session and reports the result in the GitHub job summary.
+
+The [jaas-auth](../jaas-auth/action.yml) action can be used to install Juju and
+authenticate to JAAS beforehand.
```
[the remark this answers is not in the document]
hk21702: This should be available following #133

gh-actions/infra/backup-database/tests/test_backup_database.py
```diff
@@ -0,0 +1,500 @@
...
+                "PARAMETERS_JSON": '{"nested":{"enabled":true}}',
+                "UNIT_ROLE": "any",
+                "TIMEOUT_MINUTES": "unbounded",
+            }
+        )
+        self.assertEqual(backup_target.qualified_model, "owner/model:variant")
+        self.assertEqual(backup_target.action, "custom_action")
+        self.assertEqual(backup_target.parameters, {"nested": {"enabled": True}})
+        self.assertEqual(backup_target.unit_role, "any")
+
```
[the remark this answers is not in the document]
hk21702: We now require the user to pass in the unit as well.

gh-actions/infra/backup-database/backup.py
```diff
@@ -0,0 +1,473 @@
...
+    model_owner = values.get("MODEL_OWNER", "").strip()
+    return BackupTarget(
+        application=application,
+        model=model,
+        model_owner=model_owner,
+        action=values.get("ACTION", "create-backup"),
+        parameters=parameters,
+        unit_role=values.get("UNIT_ROLE", "non-primary"),
+        timeout_minutes=values.get("TIMEOUT_MINUTES", "360"),
+    )
```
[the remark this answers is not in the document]
hk21702: We now require the user to pass in the timeout units, and we added more input enforcement.

hk21702:
Example run (with #135):
https://github.com/canonical/ubuntu-engineering-terraform-models/actions/runs/30057647433
Dry run:
https://github.com/canonical/ubuntu-engineering-terraform-models/actions/runs/30057983732

gh-actions/infra/backup-database/backup.py
```diff
@@ -0,0 +1,473 @@
...
+    """Selected Juju unit and the health information behind that choice."""
+
+    unit: str
+    requested_role: str
+    selected_role: str
+    degraded: bool
+    warning: str | None
+    workload: str
+    agent: str
+    excluded: tuple[tuple[str, tuple[str, ...]], ...]
```
CarlosNihelton: This construct is very hard to parse even for a C++ programmer :)

I believe another small dataclass is worth the cost. Something like:

```python
@dataclass(frozen=True)
class Exclusion:
    unit: str
    reasons: tuple[str, ...]
```

gh-actions/infra/backup-database/backup.py
```diff
@@ -0,0 +1,473 @@
...
+        raise ValueError("model is required")
+    model_owner = values.get("MODEL_OWNER", "").strip()
+    return BackupTarget(
+        application=application,
+        model=model,
+        model_owner=model_owner,
+        action=values.get("ACTION", "create-backup"),
+        parameters=parameters,
+        unit_role=values.get("UNIT_ROLE", "non-primary"),
+        timeout=values.get("TIMEOUT", "6h"),
```
CarlosNihelton: Can't we take those default values from granted given they come from the action inputs instead of repeating ourselves for the third time?

gh-actions/infra/backup-database/backup.py
```diff
@@ -0,0 +1,473 @@
...
+    if not eligible:
+        raise SelectionError(f"application {application} has no eligible units")
+
+    primaries = [unit for unit in parsed_units if unit["primary"]]
+    warning = None
+    degraded = False
+    if len(parsed_units) == 1:
+        selected = eligible[0]
+    elif unit_role == "any":
+        selected = eligible[0]
```
CarlosNihelton: really nitpick: don't you find and `or` expression easier to read?

```suggestion
    if len(parsed_units) == 1 or unit_role == "any":
        selected = eligible[0]
```

CarlosNihelton: Also, if there is only one parsed and eligible unit, we're selecting it without checking the requested role, what also implies we're not considering whether we should or should not mark degraded. For instance, if the input was `unit_role == "non-primary"` and we get to this point (assuming the solitaire unit is Primary), per the README that would mean a degraded fallback from requested non-primary into a Primary unit due lack of alternatives, right?

Take me with a grain of salt here because I'm not super familiar with the intricacies of the semantics of Juju.

CarlosNihelton (approved):
Awesome work. I have only minor nitpicks for your consideration, otherwise LGTM!

gh-actions/infra/backup-database/backup.py
hk21702: There are a few things that only having one eligible unit could mean. Either we only had one unit to start with, all replica units are not eligible (in error states or the like), or all units but one replica unit are not eligible.

We explicitly abort if we can't find any "Primary" unit since that's probably a situation where we should have manual intervention.

Otherwise yes, we should probably consider the case when there is only one eligible unit, but the role we want is `non-primary` to be "degraded" since that either means all replicas are non-eligible, or there's only one unit and the user should have used a different role in the first place.

gh-actions/infra/backup-database/backup.py
hk21702: Indeed, we should probably just take those defaults from the action itself and handle the case where these aren't passed as bad calls or broken implementations.

.github/workflows/test-backup-database.yaml
```diff
@@ -0,0 +1,38 @@
...
+  check-python-syntax:
+    name: py_compile
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Syntax check all Python files
+        run: |
+          while IFS= read -r file; do
```
didrocks, line 25: I always have to double check that scripts are running with `set -e` by default, but it does :)

didrocks (approved):
Good work again and nice that there are tests for the backup itself too! This looks really solid to me.

matthew-hagemann (changes requested):
I don't think we need a generalized backup solution as Ubuntu Desktop. Everyone else working on greenfield projects in the team should default to PostgreSQL and request managed DB's from IS, who need to be in charge of this operations.

For us, I think the only exception is the mediawiki operator, as we are tied to MySQL and need a solution for that. I feel that what would be better is a much smaller action, similar to [PE's backup action for PostgreSQL](https://github.com/canonical/platform-engineering-deployments/blob/main/.github/workflows/backup-github-runner-postgresql.yaml) (see comments on #133 re jaas auth).

Because we know it's a MySQL DB, we don't need to discover which actions are available or reconstruct topology by parsing  `juju status`

There is the constraint of if its a multi-node setup or not. I'd still keep it simple and in bash, or as a flag. A rough sketch based on [PE's action](https://github.com/canonical/platform-engineering-deployments/blob/main/.github/workflows/backup-github-runner-postgresql.yaml) and some idea's re not targeting the leading unit from [landscape](https://github.com/canonical/landscape-saas/blob/main/.github/actions/run-juju-ops/action.yml#L35C1-L38C21)

```yaml
inputs:
  application:
    description: "Juju application to back up"
    default: "mysql-k8s"
  no-primary:
    description: "Target a non-primary unit (required for multi-unit MySQL)"
    default: "true"
```
```sh
#!/usr/bin/env bash
set -euo pipefail

APP="${APPLICATION:-mysql-k8s}"
NO_PRIMARY="${NO_PRIMARY:-true}"
TIMEOUT="${TIMEOUT:-4h}"

# get-cluster-status must run on the leader (per charm docs).
# results.status is the InnoDB status encoded as a JSON *string* -> decode twice.
topology="$(
  juju run -m "$MODEL" "${APP}/leader" get-cluster-status --format=json \
    | jq -r '.[].results.status' \
    | jq '.defaultReplicaSet.topology'
)"
count="$(jq 'length' <<<"$topology")"

if [[ "$count" -eq 0 ]]; then
  echo "Empty cluster topology" >&2; exit 1
elif [[ "$count" -eq 1 || "$NO_PRIMARY" != "true" ]]; then
  # Single-unit cluster (primary may back up itself), or caller opted out.
  label="$(jq -r 'keys[0]' <<<"$topology")"
else
  # Multi-unit: MUST target a healthy non-primary (ONLINE + R/O).
  label="$(jq -r '
    to_entries
    | map(select(.value.status == "ONLINE" and .value.mode == "R/O"))
    | (.[0].key // empty)' <<<"$topology")"
  [[ -n "$label" ]] || { echo "No ONLINE non-primary unit available; refusing to back up the primary." >&2; exit 1; }
fi

unit="$(sed -E 's/-([0-9]+)$/\/\1/' <<<"$label")"   # mysql-k8s-0 -> mysql-k8s/0
echo "Backing up $unit (cluster member $label)"

# The charm validates health and refuses the primary in multi-unit clusters.
juju run -m "$MODEL" "$unit" create-backup --wait="$TIMEOUT"
```

I'd keep validation as dumb as possible. We would know our deployment is multi-node or not, and I'd let the MySQL operator perform validation if we are targeting the wrong unit and complain to us, instead of having too much code trying to pick the right unit to calll to.

hk21702:
> I'd let the MySQL operator perform validation 

As far as I'm aware, part of the issue is that the charm *doesn't* do any validation about if you are targeting the primary or not and if you should. The implications being that it backs up against an out of sync unit or broken unit, potentially causing a silently bad or corrupted backup, or we target the primary and cause excess load on the writing primary unit.

Once you have all that unit selection logic in place, it doesn't take that much more to genericisize it for different solutions.

matthew-hagemann:
> > I'd let the MySQL operator perform validation
> 
> As far as I'm aware, part of the issue is that the charm _doesn't_ do any validation about if you are targeting the primary or not and if you should. The implications being that it backs up against an out of sync unit or broken unit, potentially causing a silently bad or corrupted backup, or we target the primary and cause excess load on the writing primary unit.
> 
> Once you have all that unit selection logic in place, it doesn't take that much more to genericisize it for different solutions.

[lib](https://github.com/canonical/mysql-operators/blob/8.0/edge/kubernetes/lib/charms/mysql/v0/backups.py#L256-L362) does. Seems to check [you are not targeting the primary when you shouldn't](https://github.com/canonical/mysql-operators/blob/8.0/edge/kubernetes/lib/charms/mysql/v0/backups.py#L403-L404). There are also plenty of checks that you are targeting an [online unit](https://github.com/canonical/mysql-operators/blob/8.0/edge/kubernetes/lib/charms/mysql/v0/backups.py#L406C1-L407C81), on a cluster in a [valid state](https://github.com/canonical/mysql-operators/blob/8.0/edge/kubernetes/lib/charms/mysql/v0/backups.py#L374-L375) etc.

hk21702:
> > > I'd let the MySQL operator perform validation
> > 
> > As far as I'm aware, part of the issue is that the charm _doesn't_ do any validation about if you are targeting the primary or not and if you should. The implications being that it backs up against an out of sync unit or broken unit, potentially causing a silently bad or corrupted backup, or we target the primary and cause excess load on the writing primary unit.
> > 
> > Once you have all that unit selection logic in place, it doesn't take that much more to genericisize it for different solutions.
> 
> The Charm doesn't, but the [lib](https://github.com/canonical/mysql-operators/blob/8.0/edge/kubernetes/lib/charms/mysql/v0/backups.py#L256-L362) does. Seems to check [you are not targeting the primary when you shouldn't](https://github.com/canonical/mysql-operators/blob/8.0/edge/kubernetes/lib/charms/mysql/v0/backups.py#L403-L404). There are also plenty of checks that you are targeting an [online unit](https://github.com/canonical/mysql-operators/blob/8.0/edge/kubernetes/lib/charms/mysql/v0/backups.py#L406C1-L407C81), on a cluster in a [valid state](https://github.com/canonical/mysql-operators/blob/8.0/edge/kubernetes/lib/charms/mysql/v0/backups.py#L374-L375) etc.
> 
> 

Ah, I didn't see that before. 

Either way though, even with the checks to make it fail-safe, you still need to target the correct unit for the backup to run correctly. Plus, with something like this, I'd rather not risk anything when we don't need to. The alternative would be to loop through all the units until something works, and that's honestly kind of silly.

On using `get-cluster-status`, even if we assume it is always available, the amount of work to parse the topology from it versus from `juju status` isn't that different, so I don't really see a reason to restrict ourselves to assuming that it is present.

hk21702:
On simplifying to use something like a bash script, there's admittedly a good amount of logic not needed for the core operations here. But they're mostly for ease of user use, debugging, and summary generation.

Once stuff for that is added to the bash script, the size and complexity isn't that different and it's just easier to use Python, also taking advantage of unit testing with Python.

Plus, a lot of the abstractions are made in mind with the credentials backup handling done in the other PR.

hk21702:
Latest run logs: https://github.com/canonical/ubuntu-engineering-terraform-models/actions/runs/31648793128/job/94288361520?pr=711

gh-actions/infra/backup-database/backup.py
```diff
@@ -0,0 +1,693 @@
...
+        )
+
+
+def _parse_cluster_members(
+    document: Mapping[str, Any], application: str
+) -> dict[str, _ClusterMember]:
+    """Map topology member names to units, capturing memberRole and status."""
+
+    try:
+        topology = document["defaultReplicaSet"]["topology"]
```
matthew-hagemann: I'd check your cluster is eligible for backup, as you already have parsed the data required to tell. This will catch several status' that are not reflected in the charm, and thus `juju status`:

```python
replica_set = document["defaultReplicaSet"]
cluster_status = replica_set["status"]
if cluster_status not in {"OK", "OK_PARTIAL"}:
    raise SelectionError(f"cluster is not in a healthy state: {cluster_status}")
topology = replica_set["topology"]
```

gh-actions/infra/backup-database/backup.py
```diff
@@ -0,0 +1,693 @@
...
+            )
+
+        warning = None
+        if unit_role == "ANY":
+            selected_name = min(online, key=_unit_number)
+        else:
+            matches = [
+                name for name, member in online.items() if member.role == unit_role
+            ]
+            if not matches and unit_role == "SECONDARY":
```
matthew-hagemann:
```suggestion
            if not matches and unit_role == "SECONDARY" and len(members) == 1:
```

https://github.com/canonical/mysql-operators/blob/8.0/edge/kubernetes/lib/charms/mysql/v0/backups.py#L403-L404

matthew-hagemann (changes requested):
Thanks! Two nits on the gating/checks to align with what will get rejected by the backup on the mysql charm and you can merge

gh-actions/infra/backup-database/backup.py
hk21702: I'm not super sure if we need this even though there is the check in the MySQL charm. It would just be an earlier fail for the scenario where len(members) > 1 somehow, but there's probably no harm in it either.

matthew-hagemann (approved):
Thanks for the changes! LGTM 🙂

PR #135, opened 2026-07-23
feat(g/infra): Add database credential backup to Vault

hk21702:
Here, we expand on the backup-database action, adding functionality to allow for the optional backing up of database credentials to a vault, which is required for cases where the entire cluster is lost.

As the usernames needed differ not only between PostgreSQL and MySQL but also between revisions, we need to allow the user to pass in their username target.

By default, the credentials secret path is constructed as `secret/services/{model}/{application}-credentials`. The secret is created if it does not exist already.

To avoid making unnecessary writes and incrementing revisions every time the action is run, we first fetch the secrets from the vault. This is later used to compare against what the charm reports. Only if there is a mismatch do we attempt a write.

To avoid leaking credentials to the logs, we take advantage of GitHub's masking feature.

Vault credentials are provided via a token input.

The action will install the vault snap if it is not already installed, allowing for multiple calls to the action within the same job.

In a dry run scenario, we only skip write actions to the vault.

Depends on #134 

Refs:
https://canonical.com/data/mysql/docs/8.4/how-to/back-up-and-restore/migrate-a-cluster/
https://canonical.com/data/postgresql/docs/14/how-to/back-up-and-restore/migrate-a-cluster/

---


hk21702:
Example run:
https://github.com/canonical/ubuntu-engineering-terraform-models/actions/runs/30057647433
Dry run:
https://github.com/canonical/ubuntu-engineering-terraform-models/actions/runs/30057983732

didrocks:
I will let Matt and Carlos reviewing throughfully that one first. Thanks for the tests in it. I think some could have been restructed and refactored in table testing format to spot more easily the difference in behaviours, but for an action, that’s not too much of a biggie.

CarlosNihelton (approved):
I think the existing tests are different enough to justify skipping table testing for now.
I only have nitpicks and questions, no blockers.

hk21702:
Latest run logs: https://github.com/canonical/ubuntu-engineering-terraform-models/actions/runs/31648793128/job/94288361520?pr=711

matthew-hagemann (commented):
Overall LGTM! 2 Questions

hk21702:
Latest run log (non-dry-run): https://github.com/canonical/ubuntu-engineering-terraform-models/actions/runs/31741052747/job/94584239368

I also cleared out the vault for this, so this one did write credentials to the vault.

matthew-hagemann (approved):
Thanks! LGTM 🙂

